PR #1, opened 2016-02-26
Example of a hacked together build_psf

joezuntz:
Hi Mike,

I'm not sure if this fits in with the plans you had for this, but here is an example implementation of the build_psf function, which tries to run the model fitting and interpolation the way I think you intended. I had to add a getParameters method to Model to extract the numbers - not sure if you had a better way of doing this in mind?

Please do just close this immediately if it's not how you imagined this working - I just wanted to put together a minimal example that would run on the tarball you sent around.

Joe

### DIFF
--- a/piff/config.py
+++ b/piff/config.py
@@ -87,6 +87,48 @@ def read_config(file_name):
         config = yaml.load(fin.read())
     return config
 
+def build_psf(images, stars, model, interp, optics,logger):
+    """The main workhorse, which build the PSF.
+
+    :param images: A list of full exposure images
+    :param stars: A list of lists of galsim.PositionD objects with star positions.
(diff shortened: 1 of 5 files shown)


rmjarvis:
Sorry, I forgot to click the Watch button for the site (thought that was automatic for my own repos, but apparently not), so I didn't notice this or Chris's PR.

Anyway, this is great.  Thank you.  I'll probably refactor it a bit -- I want to encapsulate the stellar data a bit more than I had in the original version of the code.  But it's helpful to have a worked up version of this function help clarify how this will work.

It also needs a unit test.  I'm kind of a stickler about those.  Every bit of functionality needs unit tests.  But I can add that, unless you're keen to try to work one up.

PR #2, opened 2016-02-27
StarCatalogs

cpadavis:
3 changes:

1. Created a StarCatalog object, which keeps images and coordinates together. The StarCatalog object also has image_name and coordinate_names to keep track of what each is. For example, we can envision going from coordinates that are [x_image, y_image, ccd] to [x_focal_plane, y_focal_plane], and it might be nice to be able to check between the two. Similarly, we might represent our psf model as being in pixels, or having wcs transformations applied to them.
2. renamed readCatalog to readCoordinates, since I have created the starcatalog object (which would make it confusing) and we are currently only pulling coordinates out for that function anyways.
3. coordinates are now a flat list. In the future we should implement extracting cutouts and weights for each set of coordinates. This would entail adding some sort of keyword "size_x", "size_y" and padding.

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -45,16 +45,35 @@ def process_input(config, logger):
     input_handler = input_handler_class(**kwargs)
 
     images = input_handler.readImages()
-    stars = input_handler.readStarCatalogs()
+    coordinates = input_handler.readCoordinates()
 
-    return images, stars
+    config_star_catalog = {'images': images,
+                           'coordinates': coordinates,
+                           'image_name': input_handler.image_name,
+                           'coordinate_names': input_handler.coordinate_names}
+
+    # update the input with any extra starcatalog config keys
+    if 'star_catalog' in config:
+        config_star_catalog.update(config['star_catalog'])
+
+    # return starcatalog
+    return StarCatalog(config_star_catalog)
 
 class InputHandler(object):
     """The base class for handling inputs for building a Piff model.
 
     This is essentially an abstract base class intended to define the methods that should be
     implemented by any derived class.
+
+    :param image_name: string representing what kind of image we have (pixel,
+                       pca, moments, et cetera)
+    :param coordinate_names: a list of names indicating what the different
+                             coordinate columns mean
     """
+    def __init__(self, image_name, coordinate_names):
+        self.image_name = image_name
+        self.coordinate_names = coordinate_names
+
     @classmethod
     def parseKwargs(cls, config_input):
         """Parse the input field of a configuration dict and return the kwargs to use for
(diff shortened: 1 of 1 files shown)


rmjarvis:
Thanks Chris.  Sorry, I didn't notice this Friday -- I forgot to click the Watch button, so I didn't get an email notification of this.

I've also started working on how to encapsulate the star data, and I think we want to do this a bit differently.  I think we want an object that encapsulates all the information about an individual star.  I've been calling is StarData.  So something that would include its cutout image (vignette in SExtractor-speak), the same for the weight image, its coordinates in the image, its coordinates in the focal plane, the chip number, etc.  

Then the process_input function would return a list of StarData instances.  Then the Model classes would be able to work on a single StarData object at a time, and ask it to return the data in different formats (so far I have either an image and weight map, or a list of pixel values, weights and positions).  

The interpolators can also ask each object for the position in either coordinate system and potentially other values to use for interpolation (e.g. colors).

I'm not ready for a PR on that yet, but hopefully this week.  I'm still working up some unit tests for it.    But once I do, we can critically compare the two approaches and see which one we want to go with.

cpadavis:
Hi Mike,

So the main reason I did it as a catalog instead of the individual stars is
that I had a hard time as I wrote this figuring out what we would possibly
want to do on individual stars that we wouldn't want to apply to the whole
catalog. The main reason I did it this way is that it makes it harder to
accidentally have stars of "different kinds" -- say one star that has some
operation done to its image, and one star that doesn't. Maybe I'm just too
paranoid :)

Are you thinking, in the case of pixel data, that the individual cutouts
would live with the StarData, or that the StarData would access some stored
exposure and make cutouts on the fly?

Osprey creates an object that keeps the information on an individual star
level, like you're proposing, but doesn't do much with that extra level of
abstraction over starcatalogs. I can envision that one case where you would
want this is if you wanted to throw out stars -- which is actually a pretty
important use case, now that I think about this.

rmjarvis:
Yes, outlier rejection is one thing that gets easier with a list of stars.  You could do this with a mask column in the catalog I suppose, but deleting items from a list (or moving them to a "rejected" list) is pretty simple.  

But also, just conceptually, it seems that the way we will be using the input data is primarily one star at a time.  So we should be able to pass to a Model the information about a single star.  It seems clunkier to pass it a star catalog and the index we want it to use this time.

As for whether StarData had a copy of the cutout or a pointer to the original full image, that's an implementation detail that we can look at down the line to see if there are any performance implications.  Currently, my StarData object takes as input a galsim.Image, which can be either a fresh copy or a view into a larger image.  I'd been planning on just using the view, since it's simpler, but we can look at whether it makes more sense to make copies of the postage stamps and then delete the original.

rmjarvis:
I think this is superseded by PR #2.

PR #3, opened 2016-03-03
Implement StarData and PSF classes

rmjarvis:
I finished working up the basic infrastructure, including a StarData class to encapsulate information about each star (image, weight, position, etc.), and a PSF class which would be the full PSF solution, including a Model and an Interp.

The unit tests are still for a very simple model (Gaussian) and interpolator (Mean), and only for a single image file currently, but they now proceed through the full process from input to output and back in (reading the output file).  They also test the piffify executable, which does this whole thing according to a config file.

Very happy to get comments on the overall structure before we move on into implementing more complex models and interpolators.


joezuntz:
Hi Mike.

This looks great.  It looks like you're thinking that we'll use this library directly in the shape measurement code, just as we use galsim.des.des_psfex now to load and draw the image? And then is it straightforward to get the images at the resolution we need?

rmjarvis:
I haven't worked too much on the code for reading back in a Piff file, but yes that was kind of the plan.  So far, I can at least read back in a PSF object that was written to a file.  Making it easy to use that to draw an image of the PSF at an arbitrary position still needs some work.

rmjarvis:
BTW, I mentioned this on Slack, but might as well mention it here too.  I made a first pass at Sphinx docs for this here: http://rmjarvis.github.io/Piff/html/index.html

It mostly reads from the doc strings, but you can add any text you want in the *.rst files in the docs directory.  It's a lot nice, imo, than the doxygen docs we use for GalSim for instance.  I've been using it for TreeCorr, and really like it.

suchyta1:
I agree that sphinx is pretty awesome. So I'd second using that as the
documentation engine.

On Thu, Mar 3, 2016 at 3:28 PM, Mike Jarvis <notifications@github.com>
wrote:

> BTW, I mentioned this on Slack, but might as well mention it here too. I
> made a first pass at Sphinx docs for this here:
> http://rmjarvis.github.io/Piff/html/index.html
>
> It mostly reads from the doc strings, but you can add any text you want in
> the *.rst files in the docs directory. It's a lot nice, imo, than the
> doxygen docs we use for GalSim for instance. I've been using it for
> TreeCorr, and really like it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rmjarvis/Piff/pull/3#issuecomment-191949019>.
>



-- 
Eric Suchyta
University of Pennsylvania
Department of Physics & Astronomy
209 South 33rd Street
Philadelphia, PA 19104-6396

cbonnett:
I can second (third ?) that !
I wrote the pySkyNet documentatio in sphinx:
http://pyskynet.readthedocs.org/en/latest/

On 3 March 2016 at 21:30, Eric Suchyta <notifications@github.com> wrote:

> I agree that sphinx is pretty awesome. So I'd second using that as the
> documentation engine.
>
> On Thu, Mar 3, 2016 at 3:28 PM, Mike Jarvis <notifications@github.com>
> wrote:
>
> > BTW, I mentioned this on Slack, but might as well mention it here too. I
> > made a first pass at Sphinx docs for this here:
> > http://rmjarvis.github.io/Piff/html/index.html
> >
> > It mostly reads from the doc strings, but you can add any text you want
> in
> > the *.rst files in the docs directory. It's a lot nice, imo, than the
> > doxygen docs we use for GalSim for instance. I've been using it for
> > TreeCorr, and really like it.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/rmjarvis/Piff/pull/3#issuecomment-191949019>.
> >
>
>
>
> --
> Eric Suchyta
> University of Pennsylvania
> Department of Physics & Astronomy
> 209 South 33rd Street
> Philadelphia, PA 19104-6396
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rmjarvis/Piff/pull/3#issuecomment-191949623>.
>



-- 
=============
Christopher Bonnett
Institut de Fisica d'Altes Energies (IFAE)
Edifici Cn, Universitat Autonoma de Barcelona
E-08193 Bellaterra (Barcelona)
Spain

cpadavis:
Hi Mike,

Awesome! I like the way this looks generally. I just have a couple questions below:

Minor question: what is the desired way for people to pass in extra properties about stars (e.g. pointing)? It looks like you have a goal of making InputHandler.makeStarData not get updated between class instances, so could we possibly make it so that we can pass that information to the InputHandler and then into piff.StarData?

Another question: it appears that the following code is how we would go about creating the interpolated images, starting from a built psf object:

    psf = piff.PSF.build(stars, model, interp, logger)
    # given list of positions, interpolate parameters
    interpolated_params = psf.interp.interpolate(position)
    # set the parameters on the psf model, which gives a new model instance
    interpolated_model = psf.model.setParameters(interpolated_params)
    # create the galsim image container on 64x64 grid with some wcs system
    interpolated_image = galsim.Image(64,64, wcs=wcs)
    # use the model to draw the image
    interpolated_image = interpolated_model.drawImage(interpolated_image)

Would it be superfluous if there were a psf.interpolate_model(position) that then gives interpolated_model with the parameters already set?

rmjarvis:
I admit I've not thought very much about how we'll want to use a PSF object to draw the PSF model.  Probably the code here is woefully insufficient for that task.  I only really coded up enough on that front to write the unit tests.  

We should think about how we'll want to use a constructed PSF object (presumably typically built by reading in a file, rather than building from data) to draw the PSF at an arbitrary position.  I have a placeholder getTargetPosition in the Interp class, but that's probably not the right place, and like you say, we'll probably want to store the pointing somewhere more permanent so the user can't screw it up.  (After all, the PSF class won't work correctly if the user gives a different pointing than was used to build the PSF.)

rmjarvis:
> what is the desired way for people to pass in extra properties about stars (e.g. pointing)? It looks like you have a goal of making InputHandler.makeStarData not get updated between class instances, so could we possibly make it so that we can pass that information to the InputHandler and then into piff.StarData?

When building a PSF from data, I think the pointing should be considered an input parameter.  Either provided directly in the config file, or via header parameters like TELRA, TELDEC.  Or if it will be required (because the WCS uses celestial coords), then we could default to CRVAL1, CRVAL2 if nothing else was given, since those should always be in the header.  Then as long as we keep it consistent throughout (including keeping it with the final PSF object), we'll be ok.

rmjarvis:
I'm merging this in advance of the telecon today, since I'll be advertising it there, so just in case people click through the link.

Issue #6, opened 2016-03-16
Start work on analysis framework


rmjarvis:
@cpadavis has started work on branch psf_stats building an analysis framework for testing how good a model is on a given set of data.

This will simplify a lot of the analysis that has previously required several different pieces of software.  It would be nice to be able to
- build the PSF model
- measure some aspects of the model and stars (e.g. size, e1, e2, maybe others)
- compute some statistics of these (e.g. rho statistics)
- plot the results

all from within Piff.  For now, I would focus on the path of information flow.  Having a single choice for each step (e.g. hsm for measurements, rho1 and rho2 for stats) is fine.  Then once we settle on the design of this, we can work on adding more statistics and measurements that we might want.

Issue #5, opened 2016-03-16
Write pixel-based model class


rmjarvis:
@suchyta1 and @gbernstein have started work on the design of this class already.

This will be the first of our LinearModel classes, so we should put some thought into what aspects of this are general enough to be extracted to a middle-tier base class for other linear models.

During the hack day, we though the following would be useful features for arbitrary linear models:
1. Be able to compute the matrix that transforms from model parameters into pixel values
2. Also compute the derivatives of this wrt a centroid shift

Then the LinearModel class could use these matrices to handle the fitting and centroiding (if we are doing that -- we might want to hold nominal centers fixed) in a completely general way that will also apply to e.g. shapelet models.

PR #11, opened 2016-03-21
Psf stats

cpadavis:
Opening shots for #6  .

I decided it made sense to use the PSF class as our base class for the analysis framework. As a first step, I used the HSM algorithm in the Gaussian model combined with rho statistics as computed from TreeCorr (so you'll need TreeCorr in order to use the RhoStatistics class). Currently it computes all five rho statistics using the HSM algorithm and can make log-log plots of the resultant statistics. I appended a simple test in test_single_image in order to make sure that the rho statistics could be computed and some kind of plot generated.

I had to fix a couple things in the Gaussian model in order to make it pass a simple unit test.

I also had to create a way to generate stars from a PSF model, which means I also had to create a way to generate images from a PSF model. These are the new functions generate_image and generate_star in PSF.

Completely open to any suggestions on improving any of these bits!


piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # Read any other kwargs in the stats field
+    kwargs = stats_class.parseKwargs(config_stats)
+
+    # Build stats object
+    stats = stats_class(**kwargs)
+
+    return stats
+
+class Statistics(PSF):
```
rmjarvis: I don't think Statistics should be derived from PSF.  In object oriented programming public inheritance (the only kind in python) implies an "is a" relationship.  We don't have that here.  Statistics *is not a* PSF.  

I think Statistics should just be derived from the normal `object` class.  Then I would imagine that most Statistics classes would take a `PSF` instance as a construction argument.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        raise NotImplemented("Derived classes must define the plot function")
+
+class RhoStatistics(Statistics):
+    """Returns rho statistics using TreeCorr
+    """
+
+    def stats(self, stars, min_sep=1, max_sep=150, bin_size=0.1,
+              sky_coordinates=False, sep_units='rad',
+              ra_units='deg', dec_units='deg', logger=None):
```
rmjarvis: I think this would work better as an `__init__` method rather than a function.  Logically, I think it would make more sense to be building a `RhoStatistics` object from a `psf` object, a list of `stars` and some optional parameters.  Then rather than return things, this would set attributes: `self.logr, self.rho1`, etc. that someone could access.  Returning a bunch of numpy arrays doesn't make much sense to me.  As evidence, you don't use any of these return values in the unit test you wrote for this.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+
+    :returns: an stats instance
+    """
+    import piff
+
+    if 'stats' not in config:
+        raise ValueError("config dict has no stats field")
+    config_stats = config['stats']
+
+    if 'type' not in config_stats:
```
rmjarvis: This needs some work.  We'll at least want to allow (maybe require?) the stats field to be a list.  Not a single stat type.  Then we would want to process each statistic in turn and return a list of statistics.  Then the piffify function would probably write each one to a file I guess.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.rho = {1: rho1,
+                    2: rho2,
+                    3: rho3,
+                    4: rho4,
+                    5: rho5}
+
+        # return directly the correlation functions
+        return rho1.logr, rho1.xip, rho2.xip, rho3.xip, rho4.xip, rho5.xip
+
+    def plot(self, rho, fig=None, ax=None, logger=None, **kwargs):
```
rmjarvis: Useful to have something that returns plots so they can be displayed from within a python session.  However, more often, we will probably want to write this to a file.  So maybe have a `write()` method in the Statistics base class that will take the output a `stat.plot(...)` and write it to a a file.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+            logger.info("Getting Star Positions")
+        positions = np.array([ self.interp.getStarPosition(star) for star in stars ])
+        # generate the model stars
+        if logger:
+            logger.info("Generating Model Stars")
+        # stars_model = []
+        # for position in positions:
+        #     print(position)
+        #     stars_model.append(self.generate_star(position))
+        stars_model = np.array([ self.generate_star(position) for position in positions ])
```
rmjarvis: I think I'd rather have the image drawing be a method of `PSF`.  This isn't implemented yet, but a PSF surely should know how to make an image of itself at an arbitrary position.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+        # measure moments with Gaussian on interpolated model image
+        if logger:
+            logger.info("Measuring Model Stars")
+        shapes_model = np.array([ hsm.fitStar(star).getParameters() for star in stars_model ])
```
rmjarvis: Lots of statistics will probably want to have measurements of the shapes of stars and models.  So maybe have this functionality be in the base class?

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # define terms for the catalogs
+        ra = positions[:, 0]
+        dec = positions[:, 1]
```
rmjarvis: I don't think these are actually ra, dec.  At least in what we have so far, `getStarPosition` returns the position in field coordinates, not celestial coordinates.  So these should be u,v.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+              sky_coordinates=False, sep_units='rad',
+              ra_units='deg', dec_units='deg', logger=None):
+        """
+
+        :param stars:       A list of StarData instances.
+        :param logger:      A logger object for logging debug info. [default: None]
+        :param min_sep:     Minimum separation (in arcmin) for pairs
+        :param max_sep:     Maximum separation (in arcmin) for pairs
+        :param bin_size:    Logarithmic size of separation bins.
+        :param sky_coordinates: Are we in sky coordinates or otherwise? [default: False]
```
rmjarvis: I see.  You have this to control u,v vs ra,dec.  I don't think we want that.  I think this should always be False.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+        :param sep_units:   I think only matters if in sky_coordinates.
```
rmjarvis: sep_units can matter.   When the positions are not celestial coordinates, they can still have units.  We'll want to force that to `x_units='arcsec'`, `y_units='arcsec'`.  But `sep_units` is appropriately given by the user.  It will dictate the units used for `min_sep` and `max_sep`.  So I guess (given your doc string for them), you'll want the default value for `sep_units` to be `'arcmin'`.

piff/stats.py
```diff
@@ -0,0 +1,264 @@
...
+        # now separate the xi into positive and negative components
+        xi_neg = np.ma.masked_where(xi > 0, -xi)
+        xi_pos = np.ma.masked_where(xi < 0, xi)
+
+        if 'color' not in kwargs.keys():
+            # set color to k
+            color = 'k'
+        else:
+            # let the kwargs specify the color
+            color = kwargs.pop('color')
```
rmjarvis: Slightly simpler here:
```
if 'color' not in kwargs:
    kwargs['color'] = 'k'
...
ax.plot(r, xi_pos, linestyle='-', **kwargs)
```

tests/test_simple.py
```diff
@@ -252,6 +252,15 @@ def test_single_image():
     test_params = psf.interp.interpolate(target)
     numpy.testing.assert_almost_equal(test_params, true_params, decimal=5)
 
+    # Test that we can make rho statistics
+    stats = piff.RhoStatistics.read(psf_file)
+    logr, rho1, rho2, rho3, rho4, rho5 = stats.stats(stars, logger=logger, max_sep=5000)
```
rmjarvis: Should do some basic tests of the result here.
- check the range of separations returned
- check that the max absolute value of each rho stat is less than some value
- check that they aren't precisely zero.

tests/test_simple.py
```diff
@@ -252,6 +252,15 @@ def test_single_image():
...
+
+    # Test that we can make the rho plots
+    rho = stats.rho[1]
+    fig, ax = stats.plot(rho)
+    import matplotlib.pyplot as plt
+    plt.show()
```
rmjarvis: Don't ever have a show() command in unit tests.  They stop the test and force the user to close the window.  Better would be to write it out to a file in a `plots` directory.  Then developers can open those files if they want to make sure things look pretty.  But when doing simple regression testing, we won't need to care about them.

piff/stats.py
cpadavis: This is getting confused with my choice for inheriting the PSF class. The "generate_star" method is in piff/psf.py . I would love comments on how I went about that.

piff/stats.py
cpadavis: This was done because most of the rho statistics plots you made were in arcmin, and I wasn't sure what the units of u,v are. The other part of this is that I think treecorr calculates the distances differently if u,v is on the sphere vs Euclidean (hence the ra/dec vs x/y).

By sky_coordinates always False, we then assume u,v is Euclidean. Is that right?

piff/stats.py
rmjarvis: Right.  OK.  So here (assuming you agree with my proposal to sever this class from PSF), you would use `self.psf.generate_star(position)`.  

Although, I'm particularly fond of the name `generate_star` for that method.  I think the concept we want here is just drawing itself at a position.  Not generating a star, which is a little nonsensical.  This is an image of the PSF, not a star.  The fact that you wrap this in a StarModel is just an implementation detail so you could use the Gaussian class  so measure the shape.

But in general, I think the method in PSF should be:
```
psf.drawImage(position, offset)
```
where `offset` should default to (0,0), but would allow the user to offset the nominal center of the model from the center of the image if they want.

Then here, you could wrap that in a StarData object if you want, where the position to use in the constructor would be `image_pos = image.trueCenter() + offset`.

piff/psf.py
```diff
@@ -116,4 +116,34 @@ def read(cls, file_name, logger=None):
...
+        :param logger:      A logger object for logging debug info. [default: None]
+
+        :returns: a star
+        """
+        from .stardata import StarData
+        # TODO: not sure what to do about image? for now just make a random galsim image
+        import galsim
+        image = galsim.Image(32, 32)
+        return StarData(self.generate_image(position, image, logger=logger),
+                        galsim.PositionD(*position), logger=logger, **kwargs)
```
rmjarvis: As mentioned below, `position` is not the right thing for `image_pos` here.  `position` is the location in the field of view where you want the nominal center of the PSF to be located.  StarData takes an `image_pos` which is the location in this postage stamp image where the nominal center is located.  Based on how you wrote `generate_image`, the right value here would be `image.trueCenter()`.

But see below for an alternate proposal for how this should work.

cpadavis:
OK Next round of iterations. I think I addressed nearly all of your comments, except the one about the process_stats function. For that one I will need to read how the piffify handler works. I imagine any stats operations would go either between building the PSF model and output or after output? I'm envisioning things like plots of PSF whiskers and rho statistics

piff/psf.py
```diff
@@ -116,4 +116,28 @@ def read(cls, file_name, logger=None):
...
+        import piff
+        import galsim
+        image = galsim.Image(32, 32)
+
+        # interpolate params
+        interpolated_params = self.interp.interpolate(position)
+        # set the parameters on the psf model, which gives a new model instance
+        interpolated_model = self.model.setParameters(interpolated_params)
+        # draw on the galsim image container
+        interpolated_image = interpolated_model.drawImage(image)
```
rmjarvis: This should be
```
interpolated_image = interpolated_model.drawImage(image, offset=galsim.PositionD(*offset))
```
Then I would just return it.  I don't think we should be turning it into a StarData object.

rmjarvis:
> For that one I will need to read how the piffify handler works. I imagine any stats operations would go either between building the PSF model and output or after output? I'm envisioning things like plots of PSF whiskers and rho statistics

I think it would be worth focusing on this a bit, since I'm finding the work flow in the RhoStatistics class a bit confusing trying to understand how you expect it to work.  It seems like you would need to call the write function 5 times to get everything, which doesn't seem like a great API.  I also don't really understand the way you're passing around `fig` and `ax` parameters.  What is the use case where those should be inputs to the write function?

So I think it would be helpful to work on the top-level design.  How do we want to interact with the Statistics classes?  How would we specify this in the config file?  Also, how might we want to load a PSF from a file and run some statistics on it?  I think concrete use cases like these will help clarify what the right front-end API should be.

cpadavis:
It's confusing because I'm confused about it :)

The use cases I had in mind for passing fig and ax around would be in case one wanted to put multiple stats results onto the same figure. So for example you could be making a figure with lots of axes of psf parameters vs focal plane position. But the more I think about it the less I like it, so I have changed it with this iteration.

I have updated the piffify script so that it can take in a series of statistics. There is a simple example of this in the test_simple script.

rmjarvis:
I made a slight change to avoid using masked arrays, since I was getting runtime warnings from matplotlib not liking the masked values.  And the code is no less clear using regular masks.

rmjarvis:
Does anyone else want to code review this?  If not, I'm happy to go ahead and merge this, but if someone else is planning on giving this a look, just say so.

joezuntz:
I'll give it a quick look over.

piff/psf.py
```diff
@@ -116,4 +116,29 @@ def read(cls, file_name, logger=None):
...
 
+        :param position:    Position we are interpolating to.
+        :param offset:      Tuple; sett the nominal center of the image
+                            relative to actual center of image [Default: None]
+        :param logger:      A logger object for logging debug info. [default: None]
+
+        :returns: a galsim image
+        """
+        import galsim
+        image = galsim.Image(32, 32)
```
joezuntz: Why 32*32 in particular? Does the specific choice of this matter?

piff/stats.py
```diff
@@ -0,0 +1,364 @@
...
+
+    stats = []
+    for config_stats in config_stats_list:
+
+        if 'type' not in config_stats:
+            raise ValueError("config['stats'] has no type field")
+
+        # Get the class to use for the stats
+        # Not sure if this is what we'll always want, but it would be simple if we can make it work.
+        stats_class = eval('piff.' + config_stats.pop('type'))
```
joezuntz: it's quite hard to debug "eval" statements; I'd suggest using getattr(piff, config_stats.pop('type')). Also do you mean to modify what's in config by popping from it? Won't that cause problems if you want to run it on multiple fields?

piff/stats.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        raise NotImplemented("Derived classes must define the statistical operation!")
+
+    def plot(self, logger=None):
+        """Make your plots.
+
+        :param logger:      A logger object for logging debug info. [default: None]
+        :returns:           fig and ax
+        """
+        raise NotImplemented("Derived classes must define the plot function")
```
joezuntz: Do all statistics have to plot something?

piff/stats.py
cpadavis: This is the way @rmjarvis has been doing this in all the other files, so I just cribbed from him...

I think the usecase was that you were done with the config after processing (ie you loaded it up in a script), but I agree that you could instead do config_stats['type'] and not modify config_stats. I think in some of the other cases (like input.py), it was done so that if there were not a 'type' in the dictionary, it returned a default value:

    input_handler_class = eval('piff.Input' + config_input.pop('type','Files'))

piff/psf.py
cpadavis: Completely arbitrary! I think for our purposes it doesn't matter _that much_, and we could always make it a part of the kwargs for drawImage instead

piff/stats.py
rmjarvis: The logic on that was that other than the 'type' parameter in the config dict, the rest of the parameters there would normally be the construction kwargs for the builder.  So the `parseKwargs` function in each of the builder base classes just returns the rest of the config dict (although it can be overridden if there is more work to be done for a particular class).

Chris didn't use this paradigm for the stats classes, so maybe here it doesn't make sense.

As for whether that's a problem for repeated runs, we should probably just make a deep copy at the start of each run.  Then we can mess with the config dict as much as we want along the way without screwing things up for other runs.

rmjarvis: Actually, I checked, and we already make a deep copy at the start of piffify.  So I already anticipated this apparently.  I did make the change from eval to getattr though.  I think that's a good suggestion.

piff/psf.py
rmjarvis: I think we'll eventually want to have the model decide how big an image is appropriate to include all the information.  But let's delay that for now until we finish with #5, since I think there are some infrastructure changes that Gary and I have been discussing there that will be relevant for how to do this gracefully.

Issue #4, opened 2016-03-16
Write polynomial interpolation class


rmjarvis:
@joezuntz  has already started working on this on branch joe_polynomial.  (See also branch polynomial_interp that @danielgruen started work on, but he said he thinks all of that work is superseded by what Joe has done.)

Not much to say here.  Should take an arbitrary order for each coordinate.  Maybe options for the kind of polynomial?  Cheby, Lagrange, etc.  That might not be important though.

joezuntz:
Will do.

joezuntz:
@rmjarvis I just committed a big chunk of this with a test collection.

It turns out to be trivial to do all the other polynomials, as numpy has a whole collection of modules with very similar interfaces making Hermite, Chebyshev, Laguerre, and Legendre polynomials alongside the standard module. 

I've put a collection of tests in - they're a bit disorganized and there's definitely functionality that's not tested yet. They helped pick up a bunch of errors!  Any advice or comments v welcome.

It's currently pegged to use the same polynomial order for all the parameters and for the u and v directions. I could make it more flexible easily if there is a reason for this - my instinct is that u and v should have the same order but not necessarily all the parameters, but do you have any thoughts?

There's a few other TODO's in there also:
 - I'm using numpy arrays for the positions - you're using galsim positions elsewhere I think?
 - Probably want to normalize the coordinate ranges to keep all the coefficients the same order of magnitude.
 - Could refactor to pull out all the polynomial-specific bits to a subclass and the generic model-fitting bits in the base.

One idea: a nice milestone for Piff would be being able to run something fairly PSFEx-like on a single DES exposure file: polynomial interpolation and a pixel-wise star model.  It's a bit boring but it would be a nice test of the machinery. Does that sound good? I can add it as a github milestone if so.

rmjarvis:
Great!  Thanks Joe, I'll take a look later this weekend.  I'm traveling today and tomorrow, and I won't have much time until tomorrow night probably.

> my instinct is that u and v should have the same order but not necessarily all the parameters,

I think we should allow each parameter to have arbitrary order.  With our 2k x 4k chips, we may want to have different order for u and v when doing one chip-at-a-time interpolation.  So we should probably enable that option in general.  And more generally, we may have other parameters that we might want a different order on.  Like linear in color for instance.

> a nice milestone for Piff would be being able to run something fairly PSFEx-like on a single DES exposure file: polynomial interpolation and a pixel-wise star model.

[...]

PR #12, opened 2016-03-22
Merge polynomial interpolator branch

joezuntz:
I think we are now ready for a PR here to address issue #4 at least in part. There are a few things missing, but to get something in place for early tests of the machinery I hope this should be suitable.


piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+
+from .interp import Interp
+import numpy
+from numpy.polynomial.polynomial import polyval2d
+from numpy.polynomial.chebyshev import chebval2d
+from numpy.polynomial.legendre import legval2d
+from numpy.polynomial.laguerre import lagval2d
+from numpy.polynomial.hermite import hermval2d
+
+POLYNOMIAL_TYPES = {
```
rmjarvis: Would you mind using lower case for this (both the dict and the keys)?  Uppercase is pretty uncommon in python.  Usually reserved for constant numerical values (although `math.pi` and `numpy.pi`).  It seems like lower case would be the more typical style for this.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+}
+
+
+class Polynomial(Interp):
+    """
+    An interpolator that uses  scipy curve_fit command to fit a polynomial 
+    surface to each parameter passed in independently.
+
+    :param orders:  List/array of integers, one for each parameter 
+                    to be interpolated.
```
rmjarvis: Missing documentation of `poly_type`.  Should mention the `polynomial_types` dict for the user to see what the valid options are.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+    def __init__(self, orders, poly_type="POLY"):
+        """Create 
+        """
+        self.set_orders(orders)
+        self.set_function(poly_type)
+        self.coeffs = None
+
+    def set_orders(self, orders):
```
rmjarvis: Probably this is not a function we want the user to use directly.  Should give it a leading underscore to mark it as a "private" member function.  (Of course, nothing is actually private in Python, but this is a standard convention.)

rmjarvis: Likewise for some of the other methods below.  `set_function`, `generate_indices`, `pack_coefficients`, etc.

piff/polynomial_interp.py
joezuntz: PEP8 has upper case for module-level constants, which is what I was following here, but happy to change to whatever you prefer.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+
+    
+    def interpolationModel(self, pos, C):
+        """Generate the polynomial variation of some quantity at x and y
+        coordinates for a given coefficient matrix.
+
+        TODO: At the moment this function is expecting a numpy array of
+        shape (2,nstar) for the positions. We might want to use a galsim
+        position object instead since I think some code elsewhere in Piff
+        is expecting this.
```
rmjarvis: We should probably make our own Position class to encapsulate the "position" used for interpolation.  e.g. it may include the color or other such information.  And it should know which values constitute a regular (u,v) part of the position (and so need to use the max i+j order) vs. using having its own separate polynomial order.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+
+        """
+        #Take the u and v components (x and y in the tangent plane)
+        #as our interpolants
+        u = pos[0]
+        v = pos[1]
+        #Call the appropriate function to generate the polynomial model.
+        #By default this is numpy.polyval. Note that despite appearances
+        #this is not a method call - function is a normal python attribute
+        #that happens to be a function.
```
rmjarvis: A minor style point.  I find it hard to read comments that don't have a space after the `#`.  Would you mind adding a space?  (Here and elsewhere in this file.)

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def initialGuess(self, positions, parameter,parameter_index):
+        """Make an initial guess for a set of parameters
+        to use in the fit for your model. This is passed
+        to curve_fit as a starting point.
+
+        :param positions:  A list of positions ((u,v) in this case) of stars.
+        :param parameter:  A numpy array of the measured values of a parameter
+                           for each star
+
```
rmjarvis: `:param parameter_index:`?

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+        :param vectors:     A list of parameter vectors (numpy arrays) for each star.
+        :param logger:      A logger object for logging debug info. [default: None]
+        """
+        import scipy.optimize
+
+        #We will want to index things later, so useful
+        #to convert these to numpy arrays and transpose
+        #them to the order we need.
+        parameters = numpy.array(vectors).T
+        positions = numpy.array(pos).T
```
rmjarvis: It would be nice to log some information here.  Something like
```
if logger:
    logger.info("Fitting %d parameter vectors using polynomial type %s",len(vectors),self.poly_type)
```
Probably some extra detail at the `debug` level too would be helpful.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,361 @@
...
+            #To replace the polynomial function in this code
+            #with another model it should only be necessary to 
+            #override the methods, initial_guess, fit_function,
+            #and perhaps the pack and unpack methods
+            p0 = self.pack_coefficients(i, self.initialGuess(positions, parameter, i))
+            #print(p0)
+            # Black boxes curve fitter from scipy!
+            # We may want to look into the tolerance and other parameters
+            # of this function.
+            p,covmat=scipy.optimize.curve_fit(model, positions, parameter, p0)
```
rmjarvis: I'm a little surprised that we need a non-linear solver here.  Isn't this a straightforward matrix inversion?  Solving Ax = b -> x = (AtA)^-1 (Atb).  Is it just because the numpy polynomial functions don't give you any way to go from C -> A?  That the only interface is through that functional form `f(u,v,C)`?

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+
+    #check the packing then unpacking a 
+    packed = numpy.random.uniform(size=interp.nvariables[0])
+    unpacked = interp.unpack_coefficients(0,packed)
+    packed_test = interp.pack_coefficients(0,unpacked)
+
+    #Check that the shape is 4*4 in the unpacked (because we
+    #want space for all the terms), and that we can unpack and
+    #repack successfully.
+    assert (packed==packed_test).all()
```
rmjarvis: numpy.testing has some nicer ways to write tests involving numpy arrays.
```
numpy.testing.assert_array_equal(packed_test, packed)
```
Then for things that don't need to be exact, there are `assert_almost_equal` and similar things that tend to be useful.

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+        for j in xrange(N+1):
+            if i+j>3:
+                assert unpacked[i,j]==0.0
+                unpacked_test[i,j]=0.0
+
+    #Now do the test the other way around, checking that 
+    #we can pack and then unpack
+    packed_test_2 = interp.pack_coefficients(0,unpacked_test)
+    unpacked_test_2 = interp.unpack_coefficients(0,packed_test_2)
+    assert (unpacked_test_2 == unpacked_test).all()
```
rmjarvis: I'm confused about this test.  How could anything be non-zero here?  I'm not sure what this is supposed to be testing.

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+
+    #We also expect that if we interpolate to any point we just
+    #get the mean as well
+    for i in xrange(30):
+        p=(numpy.random.random()*10, numpy.random.random()*10)
+        v = interp.interpolate(p)
+        assert numpy.allclose(v, mean)
+
+
+def sub_poly_linear(type1):
```
rmjarvis: I think nosetests will only run functions that start with (or maybe include?) `test`.  So should change the name of this function.

rmjarvis: Oh, never mind.  This is run from the below function.  Sorry for the confusion.

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+        p=(numpy.random.random()*X, numpy.random.random()*Y)
+        assert numpy.allclose(f(p), interp.interpolate(p))
+
+def test_poly_linear():
+    for poly_type in PolynomialsTypes:
+        sub_poly_linear(poly_type)
+
+def test_poly_quadratic():
+    #This is basically the same as linear but with
+    #quadratic variation
```
rmjarvis: Any reason not to do this the same way as linear, with all the types in `PolynomialsTypes`?

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+    for i in xrange(30):
+        p=(numpy.random.random()*X, numpy.random.random()*Y)
+        assert numpy.allclose(interp.interpolate(p),interp2.interpolate(p))
+
+def test_poly_load_save():
+    for poly_type in PolynomialsTypes:
+        poly_load_save_sub(poly_type,poly_type)
+
+def test_poly_load_err():
+    from nose.tools import assert_raises
```
rmjarvis: Is this a relic?  Did you used to have this raise an exception, but now it overwrites the existing?

tests/test_poly.py
```diff
@@ -0,0 +1,287 @@
...
+    for poly_type1 in PolynomialsTypes[:]:
+        for poly_type2 in PolynomialsTypes[:]:
+            if poly_type1!=poly_type2:
+                poly_load_save_sub(poly_type1,poly_type2)
```
rmjarvis: Should put a bit at the bottom that lets you run this from the command line:

if __name__ == '__main__':
    test_poly_indexing()
    test_poly_mean()
    test_poly_linear()
    test_poly_quadratic()
    ...
```

File: piff/polynomial_interp.py (hunk 1 above)
Comment (rmjarvis) [inline piff/polynomial_interp.py]: This feels more like a "Global Variable" rather than a "Constant", in which case PEP8 recommends lower_case.  After all, there would be nothing wrong with a user adding more items to this dict if they had another kind of polynomial they function they wanted to use.  (Not that I'm a huge PEP8 stickler by any stretch -- there are a number of places where I prefer to depart from them.)

File: piff/polynomial_interp.py (hunk 8 above)
Comment (cpadavis) [inline piff/polynomial_interp.py]: I think in general it ends up being faster / more stable. This was the same thing matt did in osprey.

We should also be able to include the weight map in the interpolation fit, which should just be as simple as:

    # weights are the same shape as parameters
    sigma = np.zeros_like(weights)
    msk = weights > 0
    sigma = np.sqrt(1.0/weights[msk])
    parameter = parameter[msk]
    positions=positions[msk]
    p,covmat=scipy.optimize.curve_fit(model, positions, parameter, p0, sigma=sigma, absolute_sigma=True)

File: tests/test_poly.py (hunk 10 above)
Comment (joezuntz) [inline tests/test_poly.py]: This is testing that the polynomial coefficients with total order i+j greater than the maximum total order 3 are zero. It's testing unpacked[i,j]==0, which is not identically true since it's unpacked_test that was created just above with zeros_like, rather than unpacked; unpacked was created on line 38 by unpacked a random vector.

Comment (joezuntz) [inline tests/test_poly.py]: I will clarify with a comment.

File: tests/test_poly.py (hunk 12 above)
Comment (joezuntz) [inline tests/test_poly.py]: Good point! There was a reason originally when I was testing actual coeff values but not any more.

File: tests/test_poly.py (hunk 13 above)
Comment (joezuntz) [inline tests/test_poly.py]: Yes - deleting it now.

**Comment by joezuntz** [comment]
Okay, think I have addressed all of these comments now, apart from the curve_fit thing as it seemed like that was the way to go, and the class to encapsulate positions as that needs more work elsewhere in the code first.  Let me know if there's anything else to do.

**Comment by joezuntz** [comment]
Also, many thanks! This is a really useful exercise, as always. I wish all my code had someone reading it in this level of depth.

File: piff/polynomial_interp.py [hunk 15]
```diff
@@ -0,0 +1,389 @@
...
+        :returns:          A 2D matrix of polynomial coefficients in the form that
+                           the numpy polynomial form is expecting:
+                           p(x,y,c) = sum_{i,j} c_{ij} x^i y^j
+
+        """
+        k=0
+        n=self.orders[parameter_index]+1
+        C = numpy.zeros((n, n))
+        for k,(i,j) in enumerate(self.indices[parameter_index]):
+                C[i,j] = coeffs[k]
```
rmjarvis: Minor comment here, but this seems to have gotten extra indent spaces for some reason.

piff/polynomial_interp.py
```diff
@@ -0,0 +1,389 @@
...
+        # We will want to index things later, so useful
+        # to convert these to numpy arrays and transpose
+        # them to the order we need.
+        parameters = numpy.array(vectors).T
+        positions = numpy.array(pos).T
+       
+        # We should have the same number of parameters as number of polynomial 
+        # orders with which we were created here.
+        nparam = len(parameters)
+        assert nparam==self.nparam, """Must create Polynomial interpolator
```
rmjarvis: I think assert is the wrong thing here.  Asserts should be things that are always true if the code doesn't have bugs.  Not a way to check for user errors.  I'd rather this be
```
if nparam != self.nparam:
    raise ValueError("Must create Polynomial interpolator with ...")
```

rmjarvis:
Thanks, Joe.  Just a couple minor things now. 

So, on this branch too, would anyone else like to do a code review?

joezuntz:
Okay, cool - I've fixed those and the tests still pass. I think we'll want to revisit this whole thing later with more sophisticated models for what coefficients we have for different parameters but hopefully it's useful for now.

cpadavis:
looked good to me

PR #18, opened 2016-04-27
#5

gbernstein:
Major changes to code design, too many to list.  Highlights:

- Star class contains StarData, StarFit classes, all considered immutable.
- Model, Interpolator, PSF interfaces and routines changed.
- PixelModel is written
- BasisInterpolator for general basis-function interpolation is written, works in mode with indiv. stars giving quadratic form of chisq with parameter changes, so indiv. stars can have degenerate fits.  Polynomial basis set is coded for general selection of interpolation keys.
- Models are assumed to enforce flux normalization of the PSF and, optionally, centering at origin.
- Models should also handle the flux vs sb normalization of images.

The test/fit_des.py code demonstrates working fitting.

Things left broken or undone include:

- The existing Gaussian model will no longer work.
- The statistics will need to use new interfaces
- The existing polynomial interpolator won't work either.
- Have not written any of the read/write/parsing routines for the models and interpolators.


rmjarvis:
Great.  Thanks Gary!  I'll work on cleaning up the "broken or undone" items and ping when I think everything is working.

rmjarvis:
Update: I've finished making unit tests for all this, including making the code work via the config interface and the piffify executable.

There are still a few things about the way StarData and StarFit work that I find kind of clunky, so I'd like to change up the API of them a bit, but I'll do that on another branch to stop adding cruft to the work that Gary did here.  The code here seems to works great, as I showed at the DES meeting yesterday.  Obviously, we need to do more substantial tests to see how robust it is for a variety of input data, but that can also be deferred for now.  

So I think think this is ready for code review by anyone else who would like to take a look.

Issue #20, opened 2016-05-12
Make Star class the fundamental class for interacting with both data and fits


rmjarvis:
In #5, we added a `StarFit` class to keep track of the fitting information connected to each star.  We also added a `Star` class that doesn't do much more than hold a `StarData` and `StarFit` instance as `star.data` and `star.fit`.  This is convenient to keep track of some of the bookkeeping involved so the fit information stays connected to the relevant data even when we do things like remove outliers.

All the relevant functions are in either `StarData` or `StarFit`, and most operations (interpolation, drawing, refitting, etc.) return a new `Star` object with one or the other component replaced with an updated value.

The problem with this is that some functions still expect a `StarData` object as an input parameter.  So I'd like to clean those up to take a `Star` instead to be more uniform in the calling parameters.  

Similarly, it would be convenient if the input process would create `Star`s rather than `StarData` instances, probably with `star.fit = None`.  This would save some boiler plate in the user procedures of calling `makeStar` to turn a `StarData` into a `Star`.

Finally, I think we can move the methods from `StarData` and `StarFit` into `Star`, so the `Star` would become the fundamental class through which users would interact with the data.  Internally, we would still keep the `data` and `fit` attributes, since typically only one of these need to be changed by the various functions we have.  But that seems like an implementation detail that users shouldn't need to care about.

PR #24, opened 2016-06-21
Nearest Neighbors Interpolation and DECam Wavefronts

cpadavis:
(Trying to do the optics model pull requests in a modular fashion. There will be a PR for the interpolation, a PR for the optics model itself, and a PR for fitting the wavefront.)

In order to to do the optics model, I need to be able to pull zernike coefficients from the wavefront data. The way we've done this in the past was with a k-nearest neighbors regression to a set of reference wavefronts. Therefore I have created two new classes:

1. kNNInterp, which is a fairly generic interpolation class. It is set up so you can use it to perform nearest neighbors interpolation (via sklearn) for any starfit parameters. When initializing it, you specify attr_interp, which pulls from star.data.properties, and attr_target, which pulls from star.fit.params for the values which we interpolate to. I had some difficulty deciding how to go about the starfit parameters, since we specify that params are 1d arrays. 

2. DECamWavefront, for which you feed in a reference wavefront. When you feed stars, it looks for 'focal_x' and 'focal_y' and returns the params, which are the zernike coefficients such that params[0] = defocus.

In order to do the wavefront portion of this PR, I needed to add code that specifies how to go from pixel and ccdnum to focal plane position. This is encapsulated in pixel_to_focal and also required me adding the decaminfo object (which is in decamutil.py), which as several useful conversions from pixel to focal plane for ccdnum or extension name. That piece is pretty poorly coded, but I suspect we'll want to throw it out anyways and do any such conversions via wcs or something else a little more formal.

I also have added some unit tests and a reference wavefront from 2012 to facilitate the tests.

I am _very_ eager for your comments improving this!!


piff/decamutil.py
```diff
@@ -0,0 +1,214 @@
...
+# $Author::                                                           $:
+# $LastChangedDate::                                                  $:
+#
+# Utility methods for DECam
+#
+
+import numpy as np
+
+class decaminfo(object):
+    """ decaminfo is a class used to contain DECam geometry information and various utility routines
```
rmjarvis: In GalSim, we have survey-specific modules in sub-directories of the main `galsim` module.  Perhaps that would be a good model here as well.  So this would be `piff.des.decaminfo`  and the wavefront would be `piff.des.DECamWavefront`.

Also, as a style point, I'd recommend `DECamInfo` for this class, rather than `decaminfo`.

piff/decamutil.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    def info(self):
+        # info returns a dictionary chock full of info on the DECam geometry
+        # keyed by the CCD name
+
+        infoDict = {}
```
rmjarvis: Slightly neater here would be to have infoDict be a class attribute that gets set only once if it is not defined yet:
```
class DECamInfo(object):
    _infoDict = None

    @property
    def infoDict(self):
        if DECamInfo._infoDict is None:
            DECamInfo._infoDict = {
                'S1' : {"xCenter":  -16.908,"yCenter":-191.670, "FAflag":False, "CCDNUM":25},
                'S2' : {"xCenter":  -16.908,"yCenter":-127.780, "FAflag":False, "CCDNUM":26},
                ...
            }
        return DECamInfo._infoDict
```

piff/knn_interp.py
```diff
@@ -0,0 +1,373 @@
...
+        :param logger:      A logger object for logging debug info. [default: None]
+        :param knr_kwargs:  Arguments that will be passed to the regressor.
+        """
+
+        self.kwargs = {
+            'n_neighbors': 15,
+            'weights': 'uniform',
+            'algorithm': 'auto',
+            'p': 2,
+            }
```
rmjarvis: These are not the construction kwargs.  Although probably these parameters ought to be added to the list of parameters to `__init__`, since they look like things we'd want to have as configurable options.

piff/knn_interp.py
```diff
@@ -0,0 +1,373 @@
...
+        self.attr_interp = numpy.array(attr_interp)
+        self.attr_target = numpy.array(attr_target)
+        for target in self.attr_target:
+            self.knn[target] = KNeighborsRegressor(**self.kwargs)
+
+    def _fit(self, X, y, logger=None):
+        """Update the Neighbors Regressor with data
+
+        :param X:   The locations for interpolating. (n_samples, n_features)
+        :param y:   The target values. (n_samples, n_targets)
```
rmjarvis: We can do better than `X` and `y` for these parameters.  Give them names that mean what they are.

piff/knn_interp.py
```diff
@@ -0,0 +1,373 @@
...
+            'n_neighbors': 15,
+            'weights': 'uniform',
+            'algorithm': 'auto',
+            'p': 2,
+            }
+        self.kwargs.update(knr_kwargs)
+        self.knn = {}
+
+    def build(self, attr_interp, attr_target, logger=None):
+        from sklearn.neighbors import KNeighborsRegressor
```
rmjarvis: Need to add this to setup.py as a dependency.  Seems like a fairly innocuous one.  Took about 5 seconds to pip install, so I'm fine with adding it.  You just need to tell setup.py about it.

piff/knn_interp.py
```diff
@@ -0,0 +1,373 @@
...
+        return star_list_fitted
+
+    def writeSolution(self, fits, extname):
+        """Write the solution to a FITS binary table.
+
+        Save the knn params and the X and y arrays
+
+        :param fits:        An open fitsio.FITS object.
+        :param extname:     The name of the extension with the interp information.
+        """
```
rmjarvis: Again, these functions are unnecessary.  If you set up `self.kwargs` right, these will be properly handled by the base class implementation.

piff/knn_interp.py
```diff
@@ -0,0 +1,373 @@
...
+
+# pretty sure the correct way to do this is in the WCS framework, but let's
+# move forward first:
+def pixel_to_focal(stardata):
+    """Take stardata and add focal plane position to properties
+
+    :param stardata:    The stardata with property 'ccdnum'
+
+    :returns stardata:  New stardata with updated properties
+    """
```
rmjarvis: I'm confused about this.  Why are you interpolating in these `(focal_x, focal_y)` coordinates, rather than `(u,v)` focal plane coordinates.  The latter are almost certainly more accurate, being calculated from the WCS solutions.  And they will eventually be even more accurate once we incorporate Gary's improved WCS for Y3.  

Is there some technical reason that you need to use these instead?

aaronroodman: Mike,

This code is derived from my online AOS code which gets wavefront information for a focal plane location - I have always done this in focal plane pixel coordinates not sky coordinates.  For the F&A chips this was the only reasonable way to go, since there is no WCS solution there.   Using focal plane coordinates has been very convenient for the Science chips since alot of my work has started with the raw images and has used a private pipeline, again without the usual wcs.

Lastly, given that the wavefront PSF needs the “afterburner” I doubt extra WCS accuracy will make any difference in the results.

Aaron


On Jul 2, 2016, at 2:17 AM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


In piff/knn_interp.py<https://github.com/rmjarvis/Piff/pull/24#discussion_r69370232>:

> +        self._fit(data['X'][0], data['Y'][0])
> +        self.misalign_wavefront(data['MISALIGNMENT'][0])
> +
> +        # other attributes
> +        self.Xpixel = data['XPIXEL'][0]
> +
> +
> +# pretty sure the correct way to do this is in the WCS framework, but let's
> +# move forward first:
> +def pixel_to_focal(stardata):
> +    """Take stardata and add focal plane position to properties
> +
> +    :param stardata:    The stardata with property 'ccdnum'
> +
> +    :returns stardata:  New stardata with updated properties
> +    """


I'm confused about this. Why are you interpolating in these (focal_x, focal_y) coordinates, rather than (u,v) focal plane coordinates. The latter are almost certainly more accurate, being calculated from the WCS solutions. And they will eventually be even more accurate once we incorporate Gary's improved WCS for Y3.

Is there some technical reason that you need to use these instead?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub<https://github.com/rmjarvis/Piff/pull/24/files/02a1fef3b765a4d95d314f43082b4a18f76706d2#r69370232>, or mute the thread<https://github.com/notifications/unsubscribe/AEEa85J_6rkrWB-pjlZuxUrKaWNiikw3ks5qRbwQgaJpZM4I6RGE>.

_________________________________________________________________________
 Prof. Aaron Roodman
 Chair Dept. of Particle Physics & Astrophysics
 SLAC National Accelerator Laboratory
 Stanford University


 SLAC National Accelerator Laboratory        E-mail: roodman@slac.stanford.edu<mailto:roodman@slac.stanford.edu>
 2575 Sand Hill Rd.                                           Phone:  650-926-2705
 MS 29                                                                  Fax:  650-926-2657
 Menlo Park, CA 94025                                    URL: http://www.slac.stanford.edu/~roodman
_________________________________________________________________________

rmjarvis: But in this context, it seems like a whole lot of extra code without any benefit.  In Piff, stars already have (u,v) coordinates, which are essentially equivalent to (focal_x, focal_y), except that they are in arcsec rather than pixels and that they are potentially slightly more accurate.  So I don't see why we want to continue to use the focal_x, focal_y values here.

aaronroodman: to use the star u,v we would have to re-cast our Zernike wavefront reference meshes in terms of u,v - and will just need a good WCS solution for that data…

On Jul 4, 2016, at 7:41 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


In piff/knn_interp.py<https://github.com/rmjarvis/Piff/pull/24#discussion_r69485028>:

> +        self._fit(data['X'][0], data['Y'][0])
> +        self.misalign_wavefront(data['MISALIGNMENT'][0])
> +
> +        # other attributes
> +        self.Xpixel = data['XPIXEL'][0]
> +
> +
> +# pretty sure the correct way to do this is in the WCS framework, but let's
> +# move forward first:
> +def pixel_to_focal(stardata):
> +    """Take stardata and add focal plane position to properties
> +
> +    :param stardata:    The stardata with property 'ccdnum'
> +
> +    :returns stardata:  New stardata with updated properties
> +    """


But in this context, it seems like a whole lot of extra code without any benefit. In Piff, stars already have (u,v) coordinates, which are essentially equivalent to (focal_x, focal_y), except that they are in arcsec rather than pixels and that they are potentially slightly more accurate. So I don't see why we want to continue to use the focal_x, focal_y values here.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub<https://github.com/rmjarvis/Piff/pull/24/files/02a1fef3b765a4d95d314f43082b4a18f76706d2#r69485028>, or mute the thread<https://github.com/notifications/unsubscribe/AEEa8wa1CVYNoeXWANxS8emtL6Rx7Mo2ks5qSVPdgaJpZM4I6RGE>.

_________________________________________________________________________
 Prof. Aaron Roodman
 Chair Dept. of Particle Physics & Astrophysics
 SLAC National Accelerator Laboratory
 Stanford University


 SLAC National Accelerator Laboratory        E-mail: roodman@slac.stanford.edu<mailto:roodman@slac.stanford.edu>
 2575 Sand Hill Rd.                                           Phone:  650-926-2705
 MS 29                                                                  Fax:  650-926-2657
 Menlo Park, CA 94025                                    URL: http://www.slac.stanford.edu/~roodman
_________________________________________________________________________

cpadavis: Sorry this probably demonstrates my ignorance re: wcs coordinates. Are the u,v coordinates based on the sky, or fixed to the camera? (ie do they change with exposure?) 

I included this focal_x stuff because the wavefront model is defined in the coordinate system of the camera. There are two pieces to this. The first is that the reference wavefront (which uses the kNN interpolation to tell you what the defocus is at some point on the camera) is currently defined in terms of location on the camera, though it could be changed to be defined in terms of pixel coordinates on a given ccd (so e.g. that it loads up the kNN interpolation for ccd and zernike coefficient; currently it just loads up the kNN interpolation for zernike coefficient). The second piece is that misalignments of the camera effect the wavefront in a focal-plane dependent manner. Thus, if I were to add some decentering to the camera, there is a change in the astigmatisms that depends on where you are located on the camera.

I also wasn't sure how to implement this in the WCS, but I figured I should move forward with this and have you tell me how I go about that :)

aaronroodman: Chris etal,

Folks,

I had assumed that u,v were the “sky” version of the focal plane coordinates - ie. not RA,dec — is that correct?

Then if these are focal plane coordinates then I am also curious to know exactly how are they determined?

Chris is right to point out this issue with decenter - and I think it does suggest a reason to stay in focal plane coordinates

Aaron


On Jul 4, 2016, at 7:50 PM, Chris Davis <notifications@github.com<mailto:notifications@github.com>> wrote:


In piff/knn_interp.py<https://github.com/rmjarvis/Piff/pull/24#discussion_r69485366>:

> +        self._fit(data['X'][0], data['Y'][0])
> +        self.misalign_wavefront(data['MISALIGNMENT'][0])
> +
> +        # other attributes
> +        self.Xpixel = data['XPIXEL'][0]
> +
> +
> +# pretty sure the correct way to do this is in the WCS framework, but let's
> +# move forward first:
> +def pixel_to_focal(stardata):
> +    """Take stardata and add focal plane position to properties
> +
> +    :param stardata:    The stardata with property 'ccdnum'
> +
> +    :returns stardata:  New stardata with updated properties
> +    """


Sorry this probably demonstrates my ignorance re: wcs coordinates. Are the u,v coordinates based on the sky, or fixed to the camera? (ie do they change with exposure?)

I included this focal_x stuff because the wavefront model is defined in the coordinate system of the camera. There are two pieces to this. The first is that the reference wavefront (which uses the kNN interpolation to tell you what the defocus is at some point on the camera) is currently defined in terms of location on the camera, though it could be changed to be defined in terms of pixel coordinates on a given ccd (so e.g. that it loads up the kNN interpolation for ccd and zernike coefficient; currently it just loads up the kNN interpolation for zernike coefficient). The second piece is that misalignments of the camera effect the wavefront in a focal-plane dependent manner. Thus, if I were to add some decentering to the camera, there is a change in the astigmatisms that depends on where you are located on the camera.

I also wasn't sure how to implement this in the WCS, but I figured I should move forward with this and have you tell me how I go about that :)

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub<https://github.com/rmjarvis/Piff/pull/24/files/02a1fef3b765a4d95d314f43082b4a18f76706d2#r69485366>, or mute the thread<https://github.com/notifications/unsubscribe/AEEa8yDmOoxZUxC1yS59DekzLoi4SRJNks5qSVXrgaJpZM4I6RGE>.

_________________________________________________________________________
 Prof. Aaron Roodman
 Chair Dept. of Particle Physics & Astrophysics
 SLAC National Accelerator Laboratory
 Stanford University


 SLAC National Accelerator Laboratory        E-mail: roodman@slac.stanford.edu<mailto:roodman@slac.stanford.edu>
 2575 Sand Hill Rd.                                           Phone:  650-926-2705
 MS 29                                                                  Fax:  650-926-2657
 Menlo Park, CA 94025                                    URL: http://www.slac.stanford.edu/~roodman
_________________________________________________________________________

rmjarvis: > Then if these are focal plane coordinates then I am also curious to know exactly how are they determined?

(u,v) are coordinates in the local (Lambert) tangent plane projection of the sky around a given "pointing" direction, which would typically be the telescope pointing.  Normally (assuming the telescope pointing pointing is reported consistently), these would correspond fairly well to the focal plane coordinates you are using here.

At the pointing direction, (u,v) is defined as (0,0).  And nearby, the coordinate system preserves area, but not angles (since you can't do both).  It would be easy to switch to stereographic (preserves angles) or gnomonic (preserves great circles as lines) or Postel (preserves distances from the projection point) if one of these is preferred, since they are all implemented in the GalSim `project` function.

But I think I understand why it makes sense to keep your focal plane coordinates: these are the coordinates used for the reference solution, and it would be some work to convert that into using (u,v) coordinates.  Easier I think to leave things this way where focal_x, focal_y are used to get the reference wavefront.  

So long as (u,v) are used for the other half of the model that has the part that is likely to vary more quickly.  There, I think we want to make sure we have the right WCS solution and really use the tangent plane of the sky.

piff/knn_interp.py
cpadavis: I'm following here the example of polynomial_interp.py, or at least I think I am. I can understand why attr_interp and attr_target could be considered kwargs, but I think in the not-#20 branch, kwargs couldn't handle writing out lists or arrays, which is why they got lumped this way. If that has changed, I'll gladly put them into the kwargs dictionary.

If the X and y (now locations and targets) need to be a part of the kwargs, then I am confused by what the kwargs are supposed to be. These are the solution to the interpolator, just like the coefficients in the polynomial interp and the mean value in mean interp.

piff/knn_interp.py
cpadavis: I am okay with ditching decamutil and going to WCS coordinates. The actual reference wavefront values are reported in pixel coordinates as well as focal_x and focal_y. This means they can be converted into (u,v).

In the Wavefront model, we are concerned with terms that modify zernike coefficients based on location in the focal plane. So then at given location (focal_x, focal_y) the zernike coefficient a_i[focal_x, focal_y] = a_i[value from knn interpolant] + delta_i + theta^x_i * focal_y + theta^y_i * focal_x, where the theta and delta terms are what we fit for each exposure. Based on what you are saying, we should be able to represent the focal_x and focal_y as some sort of transform on u,v. Then, if we wanted to compare our theta parameters with my old wavefront code and aaron's active optics system, we just have to do a unit conversion.

piff/knn_interp.py
rmjarvis: Sorry, you're right. locations and targets still need to be done in this function (now `_finish_write`).  But `attr_target` and `attr_interp` are construction arguments, so they should be done by the base class `write` function via the `piff.util.write_kwargs` function.

piff/knn_interp.py
rmjarvis: The only question I guess is whether getting the pointing value where (u,v) == (0,0) is expected to be reliable enough for you using the nominal telescope pointing in the fits header.  

If so, using the regular wcs and (u,v) is probably easier in the long run.  But if we're worried about the origin of the coordinate system moving around too much because of approximations in the telescope pointing (I have no idea about this), then it might still be worth keeping this focal plane coordinate system around and using it for this purpose.

aaronroodman: I am pretty sure that the nominal pointing is not sufficient.  I’ve looked at the Blanco pointing, recently in fact, and it is only good to 10-15 arcsec, with some long tails too.




On Jul 5, 2016, at 7:14 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


In piff/knn_interp.py<https://github.com/rmjarvis/Piff/pull/24#discussion_r69611430>:

> +        self._fit(data['X'][0], data['Y'][0])
> +        self.misalign_wavefront(data['MISALIGNMENT'][0])
> +
> +        # other attributes
> +        self.Xpixel = data['XPIXEL'][0]
> +
> +
> +# pretty sure the correct way to do this is in the WCS framework, but let's
> +# move forward first:
> +def pixel_to_focal(stardata):
> +    """Take stardata and add focal plane position to properties
> +
> +    :param stardata:    The stardata with property 'ccdnum'
> +
> +    :returns stardata:  New stardata with updated properties
> +    """


The only question I guess is whether getting the pointing value where (u,v) == (0,0) is expected to be reliable enough for you using the nominal telescope pointing in the fits header.

If so, using the regular wcs and (u,v) is probably easier in the long run. But if we're worried about the origin of the coordinate system moving around too much because of approximations in the telescope pointing (I have no idea about this), then it might still be worth keeping this focal plane coordinate system around and using it for this purpose.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub<https://github.com/rmjarvis/Piff/pull/24/files/02a1fef3b765a4d95d314f43082b4a18f76706d2#r69611430>, or mute the thread<https://github.com/notifications/unsubscribe/AEEa87SKNQpvyp1QhAHJ_3QpNVLqpJ0zks5qSp8XgaJpZM4I6RGE>.

_________________________________________________________________________
 Prof. Aaron Roodman
 Chair Dept. of Particle Physics & Astrophysics
 SLAC National Accelerator Laboratory
 Stanford University


 SLAC National Accelerator Laboratory        E-mail: roodman@slac.stanford.edu<mailto:roodman@slac.stanford.edu>
 2575 Sand Hill Rd.                                           Phone:  650-926-2705
 MS 29                                                                  Fax:  650-926-2657
 Menlo Park, CA 94025                                    URL: http://www.slac.stanford.edu/~roodman
_________________________________________________________________________

cpadavis: I am really confused by all this. We don't particularly care here if the telescope is actually pointed at the spot on the sky it claims it is pointed at. We care about two things: we care that (u,v) = (0,0) is the same physical point on the focal plane, and we care that the conversion between (u,v) and (x,y) is right. If the telescope is pointing someplace other than where it thinks it is pointing, then the first condition still holds, and the second maybe doesn't. But is 10-15 arcsec going to mess up that conversion so badly?

rmjarvis: I don't know if 10-15 arcsec is good enough, but the WCS solutions are going to be right to < 1 arcsec for sure (Usually < 0.02 arcsec in fact).  So if the telescope thinks it's pointing at some other location than it really is, I think that means (u,v) = (0,0) will be at a different point on the focal plane.

However, perhaps this really means that we should have an alternate way of specifying the (u,v) = (0,0) position for des exposures.  Like use the sky coordinate corresponding to (-20,2048) on chip 35.  (Or something like this -- I'm just eyeballing it, but you probably have a better estimate that we could use.)

Then in the input field of the config we can have a few options for how to specify the pointing rather than just the ra,dec columns of the fits header (of the first file in the list...).  Like
```
input:
    ...
    pointing:
        type: DECam
```

gbernstein: It is correct that the header coordinates are off by 10-15" or more.  The astrometric-solution software will not try to update these (includes both my code and SCAMP), so if you want a focal-plane position coordinate that is more accurate than this, you should not rely on projection into the system of the header coordinates.  The strategy Mike suggests is better: choose a reference point for the optic axis, e.g. the average of the centers of S4 and N4 CCDs, and then calculate focal plane positions as the position with reference to this optic axis as (u=0,v=0).  Note also there are small rotations of the array with respect to the J2000 RA/Dec (due to precession, polar misalignment) so to get your focal-plane-invariant system (u,v) from the (RA,Dec) values given by the wcs you also want to keep track of the angle of this N4->S4 vector.

cpadavis:
Thanks Mike. I fixed a couple import bugs and also converted references of numpy to np to conform with the rest of the branch.

I think the only thing left to do here is to either implement the WCS coordinate discussion from above, or  clean up decamutil. (If we did the former, I would just discard decamutil and the focal_to_pixel function in knn_interp) Which of the two would you prefer?

rmjarvis:
I think Gary's point about the rotation might be a deal killer for using the WCS and `project`.  The tangent plane that `project` produces will always have +v in the direction of local North.  (And +u as local West.)  I think it might be hard to properly handle the (slight) rotation for different exposures.  

So I think I'm in favor of keeping the focal_x, focal_y coordinates for this purpose.

cpadavis:
OK I cleaned up decaminfo and moved it and decamwavefront into a new subdirectory in piff labeled des.

piff/des/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
+"""
+Some specialized features that specific to the DES survey and the DECam camera.
+"""
+
+from .decam_wavefront import DECamWavefront
+from .decaminfo import DECamInfo
```
rmjarvis: I moved these imports to here.  I'd rather have all the DES-specific items stay in the piff.des namespace rather than be imported into the full piff namespace.  This will be cleaner once there are multiple camera modules in piff to keep all the DES-specific things out of the main piff namespace.

tests/test_knn_interp.py
```diff
@@ -0,0 +1,231 @@
...
+    star_predicted = knn.interpolate(star_predict)
+
+    # predicted stars should find their exact partner here, so they have the same data
+    # we should still have the other interp parameter, however, so look at both!
+    for attr in attr_interp:
+        np.testing.assert_equal(star_predicted.data[attr], star_predict.data[attr])
+
+def test_yaml():
+    # using piffify executable
+    pass
```
rmjarvis: Looks like this wasn't written yet.  Should check that things work when reading this in via the config parsing.

cpadavis: this was a good thing to do. I had to move around kwargs and also make sure that write_kwargs could handle lists of strings

rmjarvis:
Before I merged this, I was doing some portability checks, running the unit tests on different systems, and I'm having a problem with the wavefront fits file you use in the DECam tests at BNL.  Perhaps @esheldon can take a look.  In particular, `fitsio` seg faults trying to read this file, but astropy can read it fine.
```
$ pwd
/astro/u/mjarvis/rmjarvis/Piff/tests/wavefront_test
$ ls
Science-20121120s1-v20i2.fits
$ python
Python 2.7.5 |Anaconda 1.8.0 (64-bit)| (default, Nov  4 2013, 15:30:26) 
[GCC 4.1.2 20080704 (Red Hat 4.1.2-54)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import fitsio
>>> f = fitsio.FITS('Science-20121120s1-v20i2.fits')
>>> f

  file: Science-20121120s1-v20i2.fits
  mode: READONLY
  extnum hdutype         hduname[v]
  0      IMAGE_HDU       
  1      BINARY_TBL      Science-20121120s1-v20i2
>>> data = f[1].read()
Segmentation fault
```
And now with astropy:
$ python
Python 2.7.5 |Anaconda 1.8.0 (64-bit)| (default, Nov  4 2013, 15:30:26) 
[GCC 4.1.2 20080704 (Red Hat 4.1.2-54)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import astropy.io.fits as pyfits
>>> f = pyfits.open('Science-20121120s1-v20i2.fits')
>>> f
[<astropy.io.fits.hdu.image.PrimaryHDU object at 0x1f05390>, <astropy.io.fits.hdu.table.BinTableHDU object at 0x1fcfe10>]
>>> data = f[1].data
>>> data
FITS_rec([ (chararray(['S', '9'], 
      dtype='|S1'), -59.8215, -69.607500000000002, -0.036317848837209302, 1.0, 20, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.021559999999999999, 1.0, -0.00064000000000000005, 0.9257200000000001, 0.01043, 1.0, -0.024059999999999998, 0.95287999999999995, -0.017440000000000001, 1.0, 0.01489, 1.0, 0.077649999999999997, 1.0, -0.0096799999999999994, 1.0, 0.0034399999999999999, 1.0, 416.99999999999977, 3795.9999999999995),
       (chararray(['S', '9'], 
      dtype='|S1'), -43.141500000000001, -72.412499999999994, -0.10775563953488372, 1.0, 20, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.018369999999999997, 1.0, -0.014069999999999999, 0.87309999999999999, 0.03986, 0.74736000000000002, -0.01485, 0.77268000000000003, 0.0049199999999999999, 1.0, 0.0098399999999999998, 1.0, 0.090999999999999998, 1.0, -0.01489, 1.0, -0.0046800000000000001, 0.81618000000000002, 1528.9999999999998, 3609.0),
       (chararray(['S', '9'], 
      dtype='|S1'), -61.831499999999998, -73.537499999999994, -0.024492965116279072, 1.0, 20, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.0068200000000000005, 1.0, 0.0073900000000000007, 1.0, 0.027539999999999999, 1.0, -0.019699999999999999, 0.75056, -0.01013, 1.0, 0.029850000000000002, 1.0, 0.058259999999999999, 1.0, -0.015169999999999999, 1.0, 0.0033600000000000001, 1.0, 282.99999999999989, 3534.0),
       ...,
       (chararray(['S', '3', '1'], 
      dtype='|S1'), -182.26050000000001, 55.947499999999998, 0.059177441860465113, 1.0, 3, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.056639999999999996, 1.0, -0.11642000000000001, 1.0, -0.0077299999999999999, 1.0, -0.11731, 1.0, 0.3024, 1.0, -0.25690999999999997, 1.0, -0.038390000000000001, 1.0, 0.013859999999999999, 0.94094999999999995, -0.056300000000000003, 1.0, 1271.9999999999995, 1517.9999999999998),
       (chararray(['S', '3', '1'], 
      dtype='|S1'), -187.94550000000001, 43.032499999999999, 0.058710988372093015, 1.0, 3, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.054960000000000002, 1.0, -0.093229999999999993, 1.0, -0.012919999999999999, 1.0, -0.13270999999999999, 1.0, 0.25253000000000003, 1.0, -0.25816999999999996, 0.61160999999999999, -0.046269999999999999, 1.0, -0.00059999999999999995, 1.0, -0.035930000000000004, 1.0, 892.99999999999932, 656.99999999999977),
       (chararray(['S', '3', '1'], 
      dtype='|S1'), -198.2355, 40.197499999999998, 0.079146453488372084, 1.0, 3, chararray(['A', 'l', 'l'], 
      dtype='|S1'), chararray(['S', 'c', 'i', 'e', 'n', 'c', 'e', '-', '2', '0', '1', '2', '1',
       '1', '2', '0', 's', '1', '-', 'v', '2', '0', 'i', '2'], 
      dtype='|S1'), 0.060920000000000002, 1.0, -0.11216, 1.0, 0.01115, 1.0, -0.1105, 1.0, 0.23615, 1.0, -0.36109999999999998, 0.78159000000000001, -0.084269999999999998, 0.62334000000000001, 0.01738, 1.0, -0.011940000000000001, 0.90264, 206.99999999999977, 467.99999999999977)], 
      dtype=(numpy.record, [('Sensor', '>i4', (2,)), ('focal_x', '>f8'), ('focal_y', '>f8'), ('z4', '>f8'), ('w4', '>f8'), ('ccdnum', '>i8'), ('expid', '>i4', (2,)), ('mesh_name', '>i4', (2,)), ('z5', '>f8'), ('w5', '>f8'), ('z6', '>f8'), ('w6', '>f8'), ('z7', '>f8'), ('w7', '>f8'), ('z8', '>f8'), ('w8', '>f8'), ('z9', '>f8'), ('w9', '>f8'), ('z10', '>f8'), ('w10', '>f8'), ('z11', '>f8'), ('w11', '>f8'), ('z14', '>f8'), ('w14', '>f8'), ('z15', '>f8'), ('w15', '>f8'), ('x', '>f8'), ('y', '>f8')]))
>>> 
```
Some of the strings seem to be wrongly stored as character arrays, but I wouldn't think that would be a problem for fitsio.  @esheldon, any idea what might be going wrong here?

esheldon:
I also see a problem with this file.

There have been seg faults with newer anaconda installs, to do with non-aligned data reading in cfitsio (numpy structured arrays are not aligned).  This may actually be a bug in the gcc,  because turning off O3 optimizations in the compilation fixed the problem in one case.  It seems there may be another case here.

Can you please open an issue on github?

esheldon:
actually looks like this was a separate bug, fixed in master

rmjarvis:
Thanks Erin.  I confirm that it works back in the original context.  I'll merge this branch now.

PR #25, opened 2016-06-21
Optics model

cpadavis:
This PR adds these features:

1. Creates a model that is the convolution of a Kolmogorov, OpticalPSF, and Gaussian, with a Shear applied at the end. This lets us model the atmosphere either as kolmogorov or as a gaussian. The shear at the end lets us also account for any common mode ellipticity. This model applies everything except the opticalPSF without reference to the stars, because they are fixed parameters of the exposure. The OpticsPSF looks for star.fit.params for an array of the zernike aberration coefficients. In other words, in this model, there is a fixed atmospheric component of a convolution of 2d gaussian with kolmogorov, and we use some sort of interpolator (like the kNN interpolator in my other pull request) to create the PSF.

2. Because the model has fixed parameters that we have to specify, I needed to add the ability to write these parameters to file when we do model.write(). This is the new functions writeParameters and readParameters, which have also modified model.write and model.read respectively.

There are also tests to make sure these things run.

Again, super eager for comments!


piff/model.py
```diff
@@ -154,29 +154,28 @@ def write(self, fits, extname):
...
-        return kwargs
+    def readParameters(self, fits, extname):
+        """Read parameters of Model from a FITS file.
+
+        By default does nothing.
+
+        :param fits:        An open fitsio.FITS object
+        :param extname:     The name of the extension to write the model information.
+        """
+        pass
```
rmjarvis: This is called `_finish_read` on #20.  And `writeParameters` is now `_finish_write`.  These complete the read or write process by writing to additional extensions other than the one with the information about the construction arguments.

piff/optical_model.py
```diff
@@ -0,0 +1,203 @@
...
+
+        :param rzero:       Atmospheric seeing. Usually in the 0.1 - 0.2 range.
+        :param g1, g2:      Shear to apply to final image. Simulates vibrational modes.
+        :param sigma:       Convolve with gaussian of size sigma.
+        :param pupil_path:  If a path is given, load up a pupil image, else
+                            make image from galsim parameters.
+        """
+
+        # catch any kwargs passed along...
+        self.kwargs = kwargs
```
rmjarvis: This isn't the intent of the `self.kwargs` attribute.  This is supposed to hold all of the construction kwargs (like `rzero`, `sigma`, `g1`, etc.).  Then the basic write function (not `_finish_write`, nee `writeSolution`) can write them out to the fits file and the read function can read them back in and know how to construct the object.  

It looks like everything here works with that paradigm without any special read/write functions.  So all you need to do to make it work is:
```
self.kwargs = {
    'rzero' : rzero,
    'g1' : g1,
    'g2' : g2,
    'sigma' : sigma,
    'pupil_path' : pupil_path,
    'lam' : lam
}

File: piff/optical_model.py [hunk 3]
```diff
@@ -0,0 +1,203 @@
...
+
+        self.lam = lam
+        self.pupil_path = pupil_path
+        OpticalPSF = {'diam': 4.274419,
```
rmjarvis: This needs to be a parameter.  Likewise for things like nstruts, obstruction, etc. that you have hardcoded for DES.  Piff functionality shouldn't be hard-coded to DES values.

Although fine to have a parameter like `optical_template = 'des'`, which would populate DES values into parameters appropriately.  Then down the line we could have other templates for other telescopes.

rmjarvis: Also a style point here: CamelCase implies a class or possibly a function, not a variable.  This should be named something like `optical_psf_kwargs`.

piff/model.py
cpadavis: Is the protocol here (since two different branches simultaneously came up with this functionality) for me to just rename the function and then when #20 gets merged in let it be overwritten?

cpadavis:
okay I think I addressed the comments with this push.

cpadavis:
I have merged with the master again. It passes the unit tests on my end

piff/optical_model.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(self, rzero=0.1, sigma=0., g1=0., g2=0., pupil_path='None', lam=500., diam=4.274419, optical_template='des', **kwargs):
+        """Initialize the Optical Model
+
+        :param rzero:               Atmospheric seeing. Usually in the 0.1 - 0.2 range.
+        :param g1, g2:              Shear to apply to final image. Simulates vibrational modes.
+        :param sigma:               Convolve with gaussian of size sigma.
+        :param pupil_path:          If a path is given, load up a pupil image, else
+                                    make image from galsim parameters referencing optical_template
+        :param lam:                 Wavelength of observations
+        :param diam:                Diameter of primary mirror. Defaults to DES value
```
rmjarvis: I don't like the default being DES-specific.  Can we couple this with the optical_template instead to have the DES value given there?

piff/optical_model.py
```diff
@@ -0,0 +1,163 @@
...
+
+class Optical(Model):
+    def __init__(self, rzero=0.1, sigma=0., g1=0., g2=0., pupil_path='None', lam=500., diam=4.274419, optical_template='des', **kwargs):
+        """Initialize the Optical Model
+
+        :param rzero:               Atmospheric seeing. Usually in the 0.1 - 0.2 range.
+        :param g1, g2:              Shear to apply to final image. Simulates vibrational modes.
+        :param sigma:               Convolve with gaussian of size sigma.
+        :param pupil_path:          If a path is given, load up a pupil image, else
+                                    make image from galsim parameters referencing optical_template
```
rmjarvis: Any reason to use `'None'` rather than `None`?  T

rmjarvis: Also, the default should be listed here in the doc string.  (Likewise for the rest of these that are missing them.)  The notation I've been trying to keep to is `[default: value]`, which matches what we decided to do for GalSim (which makes it easier for me to remember if we use the same syntax here).

cpadavis:
I believe I have addressed Mike's comments

piff/optical_model.py
```diff
@@ -0,0 +1,177 @@
...
+
+des_pupil_template = {'obscuration': 0.301 / 0.7174,
+                      'nstruts': 4,
+                      'diam': 4.274419,  # meters
+                      # aaron plays between 19 mm thick and 50 mm thick
+                      'strut_thick': 0.050 * (1462.526 / 4010.) / 2.0, # conversion factor is nebulous?!
+                      'strut_angle': 45 * galsim.degrees}
+
+class Optical(Model):
+    def __init__(self, rzero=0.1, sigma=0., g1=0., g2=0., pupil_path=None, lam=500., optical_template='des', logger=None, **kwargs):
```
rmjarvis: Why is there a `**kwargs` at the end here.  It seems that you don't use them anywhere.  It's normally bad form to accept random extra keyword arguments and just ignore them unless there is a good reason to do so.  But I don't see any reason for it here.

piff/optical_model.py
```diff
@@ -0,0 +1,177 @@
...
+
+        :param rzero:               Atmospheric seeing. Usually in the 0.1 - 0.2 range. [default: 0.1]
+        :param g1, g2:              Shear to apply to final image. Simulates vibrational modes. [default: 0]
+        :param sigma:               Convolve with gaussian of size sigma. [default: 0]
+        :param pupil_path:          If a path is given, load up a pupil image, else
+                                    make image from galsim parameters referencing optical_template [default: None]
+                                    Note: if a pupil_path is specified, then a diameter needs to also be specified
+                                    in the optical_template
+        :param lam:                 Wavelength of observations in nanometers [default: 500]
+        :param optical_template:    If no pupil plane image is given, create one from a set of templates. [default: 'des']
```
rmjarvis: I don't like having this default to 'des'.  I think that should be required to be explicit (so it defaults to None).  Then either `pupil_path` or `optical_template` would be required.

Also, I think we want to keep `diam` as a kwarg.  If the user is not requesting to use a preset template, then they need to be able to specify `diam` directly as an argument.  So it should default to None, but check that it is given if `optical_template is None`.

Finally, to make it easier to add more templates for other cameras, I would suggest having the global (module-level) variable be optical_templates, and have it be a dict indexed by the allowed values of the `optical_template` kwarg.

```
optical_templates = { 'des' : {'obscuration' : 0.301 / 0.7174, ... } }
```

Then an lsst module could have a line that would look like
```
piff.optical_model.optical_templates['lsst'] = { 'diam' : 6.5, 'obscuration' : 0.6, ... } # or whatever these are for LSST
```
And here in the init, you would just check if `optical_template` is a valid key in the `optical_templates` dict and use `optical_templates[optical_template]`.  It's a bit more maintainable this way than the way you currently do it with an explicit check for `if optical_template == 'des'`.

cpadavis:
OK I took out kwargs and added a module level optical_templates. Thanks for your patience!

tests/test_optics.py
```diff
@@ -0,0 +1,181 @@
...
+
+if __name__ == '__main__':
+    test_init()
+    test_optical()
+    test_pupil_im(pupil_path='optics_test/DECam_pupil_128.fits')
+    test_pupil_im(pupil_path='optics_test/DECam_pupil_512.fits')
+    test_kolmogorov()
+    test_shearing()
+    test_gaussian()
+    test_disk()
```
rmjarvis: Looks like this also needs a test that uses the config parsing to build the model.

cpadavis: I think I will wait for the knn to be merged in, because in going about this I discovered it currently only makes sense in the context of the decam wavefront...

maybe it's a sign that I should 'upgrade' the other components (gaussian, atmosphere, shearing) to parameter (instead of kwarg) status

rmjarvis:
@cpadavis I rewrote the `Optics.__init__` function pretty significantly [here](	1b61548).  I think we should have the same names for things here as we do in GalSim.  I'm not suggesting that those names are the best choice for everything, but having different names for e.g. `r0` and `pupil_plane_im` here will just be confusing.

I also don't think it makes much sense to have default values for `diam` and `lam`, so I made them required, although they are in the 'des' template, so they usually wouldn't need to be explicitly specified.

And finally, I made the default `r0 = None`, although again the 'des' template has `r0=0.1`.  But having this be the class default seemed to arbitrary.

I'm ready to merge this now, but you should take a look at my changes to make sure you're happy with them.

cpadavis:
Hi @rmjarvis ,

These changes are great -- thanks for writing them.

I've merged the PR since we're both satisfied.

PR #27, opened 2016-06-27
Fitting PSF Parameters

cpadavis:
Let's see if I can summarize the current model in Piff:

All Stars have Data, and some have Fit parameters. The Data includes an Images, amongst other things (Weight map, etc).

Piff has two major objects: Models and Interpolators.

Interpolators have two methods. Interpolator.interp takes a Star's Data and returns Fit parameters. Interpolator.solve takes a Star (Data and Fit and Image) and does whatever is necessary to make Interpolator.interp() work.

Models have two methods, fit and draw. Model.fit takes a Star's Data (and Image) and returns a Star with new Fit parameters. Model.draw takes a Star's Fit (and Data) and returns a Star with a model Image.

(Depending on your model, it seems natural to me that Models should also have a solve() function. I have added this into the base class, with the default solve() just passing.)

But in my Optics formalism, things are a bit more complicated. I need to fit the misalignment of the interpolator, that is, the wavefront. This wavefront takes in Star data and returns back zernike coefficients, which are fed into the Optics Model, which produces the Image. The success of my model is not Image_model - Image_data, but instead the 2nd moments (which are evaluated by the Gaussian model). Another way to put it is that the Interp.solve() method now depends on Optics Model AND the Gaussian Model. This PR is an attempt to set up a PSF formalism that can account for this.

I have created a class called HyperPSF whose goal is to fit more complicated forms like the one describe above. The particular fitting implementation is iminuit (and I have to do commit some coding sins along the way to make it work) largely because that was what I was using with the Wavefront modeling in the past. This is the chain of steps:

for each step in the minuit fitter with a list of stars and current parameters:
1. update model hyperparameters via calling new function model.update(parameters), which generically changes hyperparameters.
2. call model.solve(stars)
3. update the interpolant's hyperparameters via interpolant.update(parameters)
4. interpolant.solve(stars)
5. interpolate the stars: stars_interp = interpolant.interpolateList(stars)
6. draw the models: stars_model = [model.draw(star) for star in stars_interp)] [CPD: Could we maybe add a model.drawList(stars)?]
7. Calculate chi-square
7.1. If you give a model_comparer (say, a Gaussian), get the fit parameters of drawn stars and calculate chi-square of that
7.2. Otherwise just calculate the chi-square of the pixel map.
8. Return chi-square, think about next step.

I have implemented a very simple example of this idea in test_hyperpsf. There, I create stars with a simple 1d polynomial for their gaussian properties and I have added a sky background value to the Gaussian model. This sky background is the same for all the images, but is unknown. The hyperpsf model solves both the 1d polynomial interpolation of the Gaussian parameters with the constant sky background value.

I think this one might conflict with whatever we are trying to do on branch #20. Comments about better ways of thinking about this are always appreciated!


piff/gaussian_model.py
```diff
@@ -39,6 +39,10 @@ def fit(self, star):
         """
         import galsim
         image, weight, image_pos = star.data.getImage()
+
+        # subtract background
+        image = image - self.kwargs['background']
```
rmjarvis: I don't think we want this here.  This should be the job of the Input handler when it reads in the file.  In #20, I have it read in a `sky_col` and subtract that off, but we can add other ways to input the background value.  But I don't think we want to have each Model class have to deal with this detail.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+                  p20, p21, p22, p23, p24, p25, p26, p27, p28, p29,
+                  p30, p31, p32, p33, p34, p35, p36, p37, p38, p39,
+                  p40, p41, p42, p43, p44, p45, p46, p47, p48, p49,
+                  ):
+        # convert p to params
+        params = [p0, p1, p2, p3, p4, p5, p6, p7, p8, p9,
+                  p10, p11, p12, p13, p14, p15, p16, p17, p18, p19,
+                  p20, p21, p22, p23, p24, p25, p26, p27, p28, p29,
+                  p30, p31, p32, p33, p34, p35, p36, p37, p38, p39,
+                  p40, p41, p42, p43, p44, p45, p46, p47, p48, p49,]
```
rmjarvis: I think what you want here is:
```
def _fit_func(self, *args):
    # Pass on the function arguments to _fit_func_interior as a list
    return self._fit_func_interior(args)
```
Or even better, just have `_fit_func` be `_fit_func_interior` and use the args directly.  You don't need two functions.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+                              }
+
+        # update minuit_kwargs from set_fit_func_kwargs
+        self.minuit_kwargs.update(self._fit_func_kwargs_minuit)
+
+        # update minuit kwargs from the kwargs
+        self.minuit_kwargs.update(kwargs)
+
+        self._logger = logger
+        self._minuit = Minuit(self._fit_func, **self.minuit_kwargs)
```
rmjarvis: I think this is likely to be problematic, since here you are treating the arguments as a dict via kwargs, and below you require them to be in a particular order (`p0`,`p1`,etc. or my alternate `args` syntax).  I don't think Python guarantees kwargs to be listed in any particular order.  It's probably deterministic, so maybe it's fine.  But it's definitely bad form.  If you want the arguments to be a list, you should just treat them that way from the start, not via a `kwargs` dict.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+        :param model:                       A Model instance that defines how to model the individual PSFs
+                                            at the location of each star.
+        :param interp:                      An Interp instance that defines how to do the interpolation of the
+                                            data vectors (produced by model for each star).
+        :param model_keys, interp_keys:     A list that gives the attributes of model/interp that the params correspond to
+        :param model_comparer:              Model that gets applied to both stars and model_stars to compare success of model. If none is given, compare pixels.
+        :param model_comparer_weights:      Weights to apply to the params from model_comparer
+        :param model_init, interp_init:     A list that gives the initial values
+        :param model_error, interp_error:   A list that gives the initial step sizes
+        :param model_limit, interp_limit:   A list that gives the limits of the fit
```
rmjarvis: I changed how this function gets called a bit in #20.  All these parameters should probably be initialization parameters, not arguments of `fit`.  That way they can be specified in the configuration file.  

At least I assume we'd want these values to be configurable, right?  Or are some of these things you expect to give different values on different calls to `hyperpsf.fit`?

You can look at what I did for SimplePSF and SingleChipPSF on branch #20, although changes here to conform to that won't work until you merge branch #20 into this one.  You can either do that now to play around with it or wait until #20 is merged to master.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+        # update minuit_kwargs from set_fit_func_kwargs
+        self.minuit_kwargs.update(self._fit_func_kwargs_minuit)
+
+        # update minuit kwargs from the kwargs
+        self.minuit_kwargs.update(kwargs)
+
+        self._logger = logger
+        self._minuit = Minuit(self._fit_func, **self.minuit_kwargs)
+        if not skip_fit:
+            self._minuit.migrad()
```
rmjarvis: Also, these couple lines deserve some inline commentary to let those of us who aren't familiar with Minuit know what is going on here.  The kind of code flow I would have expected from a minimizer is:
```
solver = Minuit(func, ...)
params = solver.solve()
self.update_model_interp(params)
```
Apparently the Minuit code doesn't work this way.  I assume the Minuit constructor does both of the first two steps together.  Usually a constructor doesn't do much, so this isn't obvious here.

And it looks like you are updating the model and interp parameters along the way so they take whatever values corresponded to the last pass through `_fit_func`.  I guess Minuit guarantees that this corresponds to the minimum?

And finally, what does `migrad` do?

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+            chisq = numpy.sum(chisqs)
+
+
+        if (self.n_iter % 10 == 0) and (self.minuit_kwargs['print_level'] > 0):
+            print(self.n_iter)
+            if len(self._model_keys) > 0:
+                print(model)
+            if len(self._interp_keys) > 0:
+                print(interp)
+            print(chisq)
```
rmjarvis: These should be `logger` calls.  Not print statements.

piff/gaussian_model.py
```diff
@@ -62,21 +66,34 @@ def fit(self, star):
...
         chisq = numpy.std(image.array - model_image.array)
         dof = numpy.count_nonzero(weight.array) - 6
-
-        fit = StarFit(params, flux=flux, center=center-image_pos, chisq=chisq, dof=dof)
+        fit = StarFit(params, flux=model_star.fit.flux, center=model_star.fit.center, chisq=chisq, dof=dof)
         return Star(star.data, fit)
 
+
+    def update(self, logger=None, **kwargs):
+        self.kwargs.update(kwargs)
```
rmjarvis: I don't think you want to take any arbitrary kwargs here do you?  Only `background` is valid here.  (Although as I mentioned, I don't think we want that here, but I see why you did it for the hyper parameter tests...)  In general, you should list the valid kwargs individually so that if someone passes something that the class won't understand, they'll get a `ValueError` rather than silently updating the kwargs with something the class will just ignore.

But also, I'm not sure I like this way of keeping track of the hyper parameters.  I had been imagining these kwargs to be just used for the I/O to know how to build objects when we read them in.  I think if we're envisioning Models and Interps with hyper-parameters, we should have a struct that is designed for them specifically rather than piggy-backing on the I/O functionality.  

Maybe it's ok (since it's not invalid for these hyper-parameters to also be construction arguments), but it would require making sure that classes always look into the `self.kwargs` dict for the value of things rather than having attributes, which would be more natural.

So maybe for classes that have hyper parameters, you can have `updateHyper` which would update the hyper parameters directly.  And then not have such a function for other Models and Interps that wouldn't know what to do with them.  This seems better than silently passing as you now do for the default `update` implementation.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+                  p40, p41, p42, p43, p44, p45, p46, p47, p48, p49,]
+
+        return self._fit_func_interior(params)
+
+    def _fit_func_interior(self, params):
+        # update model and interp via params
+        if len(self._model_keys) > 0:
+            model_params = params[:len(self._model_keys)]
+            model = {self._model_keys[i]: model_params[i]
+                     for i in xrange(len(self._model_keys))}
```
rmjarvis: This would be cleaner as `dict(zip(self._model_keys, model_params))`.

piff/hyperpsf.py
```diff
@@ -0,0 +1,277 @@
...
+            interp = {self._interp_keys[i]: interp_params[i]
+                     for i in xrange(len(self._interp_keys))}
+            self.interp.update(logger=self._logger, **interp)
+            self.interp.solve(self._stars)
+
+        # interp
+        stars_interp = self.interp.interpolateList(self._stars, logger=self._logger)
+
+        # model
+        stars_model = [self.model.draw(star) for star in stars_interp]
```
rmjarvis: No problem with adding a `drawList` function analogous to `interpolateList` for this.

piff/model.py
```diff
@@ -110,6 +110,26 @@ def fit(self, star):
...
 
+    def solve(self, star_list, logger=None):
+        """Solve for any model coefficients given some data.
+
+        :param star_list:   A list of Star instances to fit model parameters with
+        :param logger:      A logger object for logging debug info. [default: None]
+
+        By default does NOTHING since you may have no hyperparameters to put in
+        """
+        pass
```
rmjarvis: I'm not sure I understand what this is supposed to do when it's not doing nothing.  Perhaps it will make more sense when coupled to your OpticalPSF classes and I can see a real world non-trivial example.

piff/hyperpsf.py
cpadavis: This is a consequence of the way iminuit works. It _needs_ to have named parameters, and I couldn't figure out how to specify the names of a function's variable dynamically.

piff/hyperpsf.py
cpadavis: A lot of the problems here are coming from the fact that iminuit needs to have function variables be named. Since I couldn't figure out how to dynamically define the fit function to be like function(sky_background, atmosphere), etc, I use _set_fit_func_kwargs to create the dictionaries that map variables to the p's (so here sky_background to p0, atmosphere to p1). This is deterministic.

The Minuit object is initialized to set things like the limits and initial guesses of the parameters.

piff/hyperpsf.py
cpadavis: I will probably wait for #20 to be merged before I fix here. Agree that they should be initialization parameters, not arguments of fit. The only parameter I expect one might call on different calls is skip_fit, for debugging purposes.

rmjarvis: That's fine.  If you want to comment on anything in #20, please do.  Maybe we can plan to merge that this week sometime, and then see if this all makes sense with that code flow.

piff/model.py
cpadavis: The optics model as implemented won't actually have this either, but I think it's a feature of all models, just like with interpolators: if an interpolator sometimes needs to learn the hyperparameters to go from positions to fit parameters, it seems reasonable to me that a model might need to learn hyperparameters to either go from images to fit parameters or fit parameters to images.

Here is a non-trivial example: PCA of the images. You can't determine the PC's in advance; you have to feed it some set of images for it to make that.

Another example is if we want to interpolate over whitened stars (ie mean subtracted and/or variance normalized) or apply a model to whitened stars. I am not sure why we would want to do this, but in that case you would need to 'solve' for the mean and/or variance of the pixels so that you can properly fit and draw.

piff/hyperpsf.py
rmjarvis: I looked into this a little bit, and I at least see where you are coming from here.  However, I think iminuit has ways around this problem.  The [Tutorial](http://nbviewer.jupyter.org/github/iminuit/iminuit/blob/master/tutorial/tutorial.ipynb#Faking-a-function-signature) describes ways to fake a function signature.  In particular, it looks like [`forced_parameters`](http://nbviewer.jupyter.org/github/iminuit/iminuit/blob/master/tutorial/tutorial.ipynb#Last-Resort:-Forcing-function-signature) might do what we want.

Please look into the solutions described here and see if you can get one of them to work.  The current solution is too ugly for me to want to leave it this way.

cpadavis:
I'm going to close this since the branch I've been working on the wavefront model in is called wavefront. It currently just needs me to get around to finishing the unit tests and merging master, and then I will PR it.

rmjarvis:
OK.  I was going to ping you about this to see if this should be closed.  :)

cpadavis:
You basically did with the push :)

Issue #26, opened 2016-06-22
glob.glob() can return bizarre ordering


erykoff:
Hey, stackexchange has the (unsatisfying) answer:
http://stackoverflow.com/questions/6773584/how-is-pythons-glob-glob-ordered

Apparently, it's order in the filesystem, whatever that may be; the order they were written to disk?  But for two globs of files, they are by no means guaranteed to have the same matched ordering.  

(I think sorting might fix this but would require some testing)

esheldon:
I always do the following
```
flist=glob(pattern)
flist.sort()
```
note sort is in place, so this will *not* work
```
flist=glob(pattern).sort()
```

rmjarvis:
Or
```
flist = sorted(glob(pattern))
```
But the point remains that it would be helpful to have other ways to specify the list of image and catalog files to make sure they match up 1-1.

esheldon:
FYI, sorted() returns an iterator, not a list.

rmjarvis:
I don't think so...
```
>>> type(sorted(glob.glob('*.fits')))
<type 'list'>
```
or in Python 3.4 (where I thought maybe they changed the nature of this function
```
>>> type(sorted(glob.glob('*.fits')))
<class 'list'>
```

esheldon:
I was clearly confused

erykoff:
I did not know about the sorted() thingy.  I always did it the way Erin did.

In any event, as Mike has said, the original point still stands that we
need to guarantee 1-1 matching.  It may be that simply using sorted() does
the trick if you have filenames that are all the same except for some
obviously sortable index, which is the "expected" behavior.

esheldon:
In my experience it is best to determine such things algorithmically.  For example, if I'm processing an exposure I can generate the file names for the ccds in the exposure using a simple algorithm based on the DESDM run identifier.  The base name is known and the suffix is based on ccd number.

I recommend doing that by default. e.g. specify an exposure you want to process by name or id, and the code looks for it in the usual place `($DESDATA/OPS/red/...)` and does its thing.

This has the advantage that if a file is missing you will notice it; you generated the file list and you know what should be there, if the file is missing the code will crash.  Also there are usually auxilliary files you will need to use that will be generated using the same algorithms, and don't need to be determined from the input file list.

Then if you want to be able to do something different, perhaps for testing, that can be an alternative method, e.g. specifying an exact list of files, or a pattern as in the above config.

rmjarvis:
[...]

Issue #13, opened 2016-04-01
Implement outlier rejection


rmjarvis:
We need some machinery to handle finding outliers in the interpolation and removing them (and redoing the interpolation).

In the config structure, I think this can be a parameter within the interp field, but in the code, the functionality will be separate from any particular Interpolation class, working with the residuals between the measurements and the interpolated values.

I volunteer to do this, and I plan to implement a few different options.  This [article](http://www.r-bloggers.com/finding-outliers-in-numerical-data/) has a good discussion of the typical kinds of outlier rejection schemes that are usually employed.  I think we should implement their first three options (based on sigma, median absolute deviation, and quartiles), and probably follow their lead in making the MAD scheme the default.

gbernstein:
There would be a few options to consider about rejecting bad data:

1. As you imply above, reject a star if one or more of its best-fit PSF parameters is an outlier from the interpolated pattern.
2. Reject a star if the PSF, interpolated to its location, yields a poor chi-squared (or, more generally, log likelihood) when fit to the data.
3. Reject a star (or some pixels) if the interpolated PSF is a poor fit to one or more of its pixels.

For the latter two options it might be useful to have the Model be able to return the chisq when the model is applied to some StarData.

RobertLuptonTheGood:
I've written code like this, and it can be surprisingly hard to get it to behave.  If you have only a few stars in some corner of a chip with a significant amount of distortion, it's easy to clip them away with an initial guess.   Maybe you've already got beyond this?

rmjarvis:
Right.  I also have code that did this, and I had to set the outlier threshold pretty high to not throw away real data, since my interpolation couldn't track the data terribly well.  (This was using the BTC camera, where one chip had a significant warp that meant the PSF wasn't all that smooth.)

I also kept track of the bounding polygon of the stars that survived outlier rejection and made that the confidence region for the PSF.  Any galaxy outside that polygon was rejected from my shear catalog, since I considered the PSF there suspect.  That way whenever I did throw out stars in the corners, I at least didn't trust my extrapolation there.

Not sure if we need to be that draconian here, but we'll at least want to make some diagnostics to keep track of where the outliers are, so we can investigate if they are systematically in corners and such.

gbernstein:
[...]

PR #28, opened 2016-07-01
#20 Lots of updates to front end API

rmjarvis:
This is a big PR with lots of changes to the front-end and mid-level API as well as a few new features.  I was mostly working towards getting [this config file](https://github.com/rmjarvis/Piff/blob/%2320/examples/des.yaml) working for running the existing pixel-based model and polynomial interpolation on a full DES field.

API changes:
- Made Star the principal class for dealing with information about each star (either real star or a model of the star at some location).  It still keeps `data` and `fit` attributes, since most things that will make a new star only change one or the other of these.  But that's now more of an implementation detail, rather than something the user needs to know about.  This is issue #20.
- Made the PSF class a base class in the same way we do the others (Model, Interp, Stats, etc.), so we can have several different "type"s of PSFs specifiable in the config file.  I needed this for the "SingleChip" PSF class described below.  And Chri's HyperPSF (cf. #27) will fit into this structure as well.
- Moved the `stats` outputs to a subfield of the `output` field in the config file.  This seems more natural to me, both in implementation and structure of the config file, than having each stat field have its own `output` subfield.
- Moved the `model` and `interp` fields to subfields of `psf`, which are now a detail about the default `SimplePSF` and also the new `SingleChipPSF`.  I think this addresses most of Chris's concerns about HyperPSF working differently.  That class would not need to have `model` and `interp` as components.
- Changed the name of PixelModel to PixelGrid, which I think is slightly more descriptive as to what it is doing.

New features:
- New SingleChipPSF type that has a separate SimplePSF (what used to be the only kind of PSF) for each chip.  It writes out a single output file with the solutions for all chips in the exposure, but each one is an independent solution for that chip.
- Added more ways of specifying the list of input files.  This resolves issue #26.
- Added outlier rejection as a new subfield of `psf`, and implemented a simple rejection scheme based on the computed chisq for each star (called ChisqOutliers).  This resolves issue #13, although it is likely that we will want additional outlier schemes as well.
- Started work on a Mean Absolute Deviation (MAD) outlier rejection algorithm, but I realized it wasn't what we wanted for PixelGrid, so I abandoned it.  But the start of an implementation is there.
- Ported over my version of the rho stats plot, which I think has better handling of positive vs negative values.  This is based on the code I used to make the rho plots for the DES SV paper.

Bug fixed:
- I ran into oscillatory behavior in the PixelGrid.reflux function that meant it never converged.  Basically the centroid adjustment would change which pixels would be used for the fit, and then the centroid would move back.  Repeat.  I fixed this by turning off the centroid fitting whenever the chisq increases from the value in the previous iteration.

Results:
- The example config file runs to completion without errors on this one exposure.
- It is quite a bit slower than PSFEx.  It took 310 minutes on my laptop, which is an average of about 5 min per chip.  PSFEx typically takes around 1 minute per chip.  So there is probably still a fair bit of optimization potential here.  Maybe by pushing some of the linear algebra in particular down to a C++ layer.  Leaving that for another issue...
- The rho stats for this image doesn't look great (see below), although similar in magnitude to what PSFEx gives for this same exposure, so it's encouraging enough for me to try it on a larger test bed to see if we are doing any better when we have better statistics.

decam_00241238_rho


rmjarvis:
For reference, here are the corresponding plots for PSFEx.

[rho1_all_r.pdf](https://github.com/rmjarvis/Piff/files/344033/rho1_all_r.pdf)
[rho2_all_r.pdf](https://github.com/rmjarvis/Piff/files/344034/rho2_all_r.pdf)

Looks like the Piff rho1 is maybe a bit worse at small scales, but rho2 is better at most scales.  Hard to say though, since so many error bars here are consistent with zero.

piff/basis_interp.py
```diff
@@ -46,51 +46,53 @@ def __init__(self):
...
+        This class will initialize everything to have constant PSF parameter vector taken
+        from the first Star in the list.
+
+        :param stars:       A list of Star instances to use to initialize.
         :param logger:      A logger object for logging debug info. [default: None]
 
         :returns:           A new list of Stars which have their parameters initialized.
         """
+        c = stars[0].fit.params.copy()
+        self.q = c[:,np.newaxis] * self.constant(1.)[np.newaxis,:]
```
cpadavis: So all classes based on BasisInterp must have a `constant' function? Maybe a placeholder function should be put in so we know what it needs to be.

(I see this is already in the master branch. I apologize for not catching this earlier!)

rmjarvis: Right.  This is Gary's code -- I just moved things around a bit (and regularized to using `np.` since we had had a mix of `np` and `numpy`.  

The constant function just makes the appropriate polynomial (at least in BasisPolynomial it is a polynomial) that corresponds to a constant value in (u,v).  I could add a base class method that raises NotImplemented with an appropriate message saying that you need to use a concrete derived class, not the abstract base class.

piff/input.py
```diff
@@ -262,6 +372,80 @@ def __init__(self, images, cats,
         self.gain = gain
         self.pointing = None
```
cpadavis: it looks like ra / dec are not used at all for setting the pointing?

piff/pixelgrid.py
```diff
@@ -618,44 +619,72 @@ def reflux(self, star, fit_center=True):
...
                                                center = center,
                                                chisq = chisq,
                                                worst_chisq = worst_chisq,
-                                               dof = np.count_nonzero(weight) - self._constraints,
+                                               dof = dof,
                                                alpha = star.fit.alpha,
                                                beta = star.fit.beta))
+            # If chisq went up, turn off centering.  There are a number of failure modes
+            # to this algorithm that can lead to oscillatory behavior, so if we start doing
+            # that, just turn off the centering for the next iteration.
```
cpadavis: It turns it off for the rest of the fit, not just the next iteration, right? Do we also want to undo the last shift in the center?

piff/simplepsf.py
```diff
@@ -0,0 +1,223 @@
...
+            logger.debug("Initializing interpolator")
+        self.stars = self.interp.initialize(self.stars, logger=logger)
+
+        # For basis models, we can compute a quadratic form for chisq, and if we are using
+        # a basis interpolator, then we can use it.  It's kind of ugly to query this, but
+        # the double dispatch makes it tricky to implement this with class heirarchy, so for
+        # now we just check if we have all the required parts to use the quadratic form
+        quadratic_chisq = hasattr(self.model, 'chisq') and self.interp.degenerate_points
+
+        # Begin iterations.  Very simple convergence criterion right now.
```
cpadavis: If the PSF model doesn't have refluxing and outlier rejection, then the below for loop will run either 1 time (fit is below threshold) or max_iterations times (fit is above threshold but there is no way to change the fit since there is no refluxing or outlier rejection). In that case, since the end result is always the same, doesn't it make sense to just run the loop once?

piff/stats.py
```diff
@@ -20,184 +20,227 @@
...
+        :g1_model   The g1 component of the PSF model at these locations.
+        :g2_model   The g2 component of the PSF model at these locations.
+        :dT:        The size residual, T - T_model
+        :dg1:       The g1 residual, g1 - g1_model
+        :dg2:       The g2 residual, g2 - g2_model
+    """
+    def __init__(self, bins_size=10, bins_shape=10, file_name=None, logger=None):
         """
+        :param bins_size:   Number of bins for size histograms. [default: 10]
+        :param bins_shape:  Number of bins for shape histograms. [default: 10]
```
cpadavis: It is worth pointing out (now that I think about this) that these can actually also be numpy arrays specifying bin edges as per plt.hist

cpadavis:
OK I went through the code. Aside from some very few comments this looks great to me.

piff/pixelgrid.py
rmjarvis: > Do we also want to undo the last shift in the center?

I'd thought about doing that but didn't bother.  I guess it's worth doing though.  When this failure mode hits, I think there is a fundamental uncertainty in what the right centroid is that is of order the difference between the two values.  But since the previous value had the smaller chisq, I guess it's worth reverting back to that.  Done here: 097ce7b

piff/simplepsf.py
rmjarvis: > If the PSF model doesn't have refluxing and outlier rejection, then the below for loop will run either 1 time (fit is below threshold) or max_iterations times (fit is above threshold but there is no way to change the fit since there is no refluxing or outlier rejection).

Not quite.  The test is on `oldchisq - chisq`.  So if there are no changes, this will be 0, which will test as less than the (positive) `chisq_threshold * dof`.  So it will stop on the second iteration.

I think models _should_ always have a `reflux` function.  Certainly, all relatively sophisticated models will want to have this to handle the case of missing data (masked pixels).  The best estimate for those missing pixels comes from the interpolated model.  

The current check about `if hasattr(self.model, 'reflux')` is kind of a hack to avoid the issue for now, since I didn't implement something appropriate for `Gaussian`. And in practice we'll almost always want to use some kind of outlier rejection.  So I think I'm ok with not having an explicit check for the case where there is no mechanism for chisq to decrease.  It's a pretty minor inefficiency for an unusual use case.

piff/stats.py
rmjarvis: Done. 6917d11

piff/basis_interp.py
rmjarvis: Done. 	088c94f

rmjarvis:
Thanks Chris.  I think I've addressed everything.  I'll plan to merge this tomorrow unless anyone else would like to take a look through the code first.

PR #32, opened 2016-07-21
Two dimensional colormaps of PSF Fits

cpadavis:
In doing my wavefront analyses, I have found it useful to be able to produce colormaps of PSF fits, where the x and y axes are the location of the star (in u, v here) and the color is the value of the parameter (say, difference between star's and model of star's sigma). I had also found that these plots tended to be easier to interpret if, instead of looking at the mean value in a given u-v voxel, I looked at the median value.

I have written a new stats package that does exactly this with just numpy. It is meant to be easy to extend. If you wanted, say, to find the standard deviation in the size in a voxel using numpy's std function, you could pass reducing_function='np.std'. Or, if you wanted to find the maximum value, you could pass reducing_function='max'. By default it finds the median.

You can also customize the number of bins in the u and v directions. If we wanted, it would be easy to create a class for DECam where the bins are, say, some fixed number of regions per ccd.

It can also handle regions with no objects. This makes plotting the entire DECam focal plane painless.

There is a basic unit test in test_twodstats.py, and I have extended test_simple to include twodstats for the purposes of testing piffify and yaml.

Attached is an example plot, where I randomly generated stars with a linear increase in sigma and g2 and a parabolic increase in g1. I fit with a linear 1d polynomial. I also exclude a portion of the center, to demonstrate the masking.

twodstatsstats


tests/test_simple.py
```diff
@@ -315,6 +315,19 @@ def test_single_image():
             'type': 'Rho',
             'file_name': rho_psf_file
         },
+        {
+            'type': 'TwoDHist',
+            'file_name': os.path.join('output', 'simple_psf_twodhiststats.png'),
```
rmjarvis: In my portability tests recently, I found that .png files had issues on one or two systems.  Something about the libpng library.  I don't know whether it is a bug in matplotlib or with the system's libpng library or what exactly, but using .pdf files (with the 'Agg' backend) worked everywhere, so I think we should stick with that for tests such as these.

If people have a working png parser on their system, obviously that would still work for them.  But I don't want to have my unit tests fail because of some png problem with matplotlib.

tests/test_twodstats.py
```diff
@@ -0,0 +1,92 @@
...
+    psf.fit(stars, None, None)
+
+    stats = piff.TwoDHistStats(number_bins_u=21, number_bins_v=21, reducing_function='np.mean')
+    stats.compute(psf, stars)
+    # Test the plotting and writing
+    twodstats_file = os.path.join('output','twodstatsstats.png')
+    stats.write(twodstats_file)
+    # fig, axs = stats.plot()
+    import matplotlib.pyplot as plt
+    plt.show()
```
rmjarvis: This is not a valid unit test. 

1. There is no testing here. You need some actual asserts.  So far this just checks that there are no syntax errors in the code, which is a pretty low bar for a test.  
2. You can't have the test pop up a plot on the screen.  For one, people may not be running in an environment where that is even possible.  And moreover, we don't want to have to close this plot window every time we run nosetests.

rmjarvis:
This seems broadly fine to me.  (cf. a couple of inline comments)  I don't personally find g1 g2 plots like this to be often very useful.  Whisker plots tend to be much more informative, imho.  But I'm fine with having this is the piff stats arsenal.

tests/test_twodstats.py
cpadavis: On point 1: I'm a bit at a loss for what I should do for a plotting PR. Do I make a reference image and assert that the two images are the same?

rmjarvis: > On point 1: I'm a bit at a loss for what I should do for a plotting PR. Do I make a reference image and assert that the two images are the same?

Test that compute is computing the right thing.  The image is just one representation of the statistic, but the raw numbers are amenable to testing.

cpadavis:
Luckily whisker plots are easy to add. So I added them, following the style you used in the shear paper.

I also added a few assert statements in test_twodstats to make sure the averaging in the bins was working as expected.

PR #34, opened 2016-07-25
Med. Abs. Dev. Outliers, Multiple Exposures, etc.

dchaiken:
Added outlier rejection based on Median Absolute Deviation, new psf class that can run piffify on multiple exposures simultaneously, stats subclass that creates scatterplots of observed vs. modeled T, g1, and g2, and a few other small fixes. This also includes Chris' code to create 2d histograms for the data.


cpadavis:
Hi Daniel, All,

For those who are not Daniel, Daniel is Aaron's summer student, and is going to be a freshman this fall. We've been teaching him about weak lensing, PSFs, data analysis, &c using Piff.

The primary thing Daniel has added here is the multiexposure PSF, which is like a singlechipPSF, but now upgraded to also include multiple exposures. The first context we are using it in is accumulating statistics over multiple exposures. I am not sure if this is desirable for everybody else, nor am I sure that this is the way we want to be able to make (e.g.) rho statistics for a testbed. It was, however, a good pedagogical exercise for Daniel.

The other things Daniel has added are a MAD outliers class, and a stats module that plots the 2d scatter of the shapes of the psf and stars.

rmjarvis:
I'm sorry, but this PR is kind of a mess.  This is going to sound a bit harsh, and I understand that this is a college freshman, but this really needs a lot of work before I can even give it a proper review.  Here are some high-level points about the work that needs to happen.

1. Some of the files changed involve mostly a retabbing to 8-space tabs.  (a) I'm not going to accept 8-space tabs, so fix that.  And (b) I can't actually tell what was substantively changed in these files.
2. Some of the new stuff also has 8-space tabs.  Fix that.
3. The new MAD outlier code should have been put into what I already started on that, rather than making a new class.
4. There is no unit tests of the MAD outlier rejection algorithm.  You need some.
5. The changes to input.py broke the existing unit tests.  Make sure nosetests pass before you submit a pull request.
6. The addition of test stars is likely something we want to have, although I'm not a huge fan of the current implementation.  Especially the bit you added to config.py with warnings and hardcoded use of a Gaussian.  This seems all wrong.  I haven't looks at much of the rest of the tstars implementation yet (pending tabbing fix).
7. There are no unit tests of the new tests stars functionality.
8. Building a PSF from multiple exposures would require some significan API design to consider how best to implement it.  We'll need to think about how best to indicate such a thing in the config file, and a strawman implementation should at least do something with multiple exposures rather than just a "SingleExpPSF" that is currently the only thing in multiexp.py.  (Which looks suspiciously like a near verbatim copy of SingleChipPSF.)  This is a big job and probably would benefit from some significant collaborative planning sessions to flesh out how it should work.  My suggestion here is to abandon this part of the PR until we have a real use case that will require such functionality.  In any case, the new functionality should be a separate PR from the other pieces you are submitting here, since it is a pretty different topic.
9. Probably the other two topics (MADOutlier and test stars) should be split up into two PRs, since they aren't really related.

Please let me know when you think you have something ready for me to code review.  If you want to close this PR and reopen a new one (or two separate ones), that would be fine too.

Finally, @dchaiken I've added you as a developer for Piff, so if you want, you can work on a branch in the Piff directory rather than a fork.  The fork is fine of course, so it's up to whatever your preference is.  But I tend to find it easier to work on a branch where any of us can directly push changes.

Issue #38, opened 2016-09-12
Kolmogorov model


jmeyers314:
Add a Kolmogorov model, which I think will look a lot like a Gaussian model in implementation.  Combined with a good interpolator, this could be useful for modeling the atmospheric part of a PSF.

Before I start, any recommendations for the `.fit` method?  `import lmfit`?  Rescale HSM result?  ...?

rmjarvis:
There are a couple of things that the Gaussian model doesn't do that @gbernstein added with the PixelGrid model.  

1. We'd like to have the option of fixing the center (nominally the centroid).  Especially for something like LSST, the location of the star will be much better measured from the stack fit of a star than from any individual exposure.  But more importantly, this allows for the possibility of including small centroid shifts that aren't captured by the WCS as part of the PSF.  If these shifts are due to atmospheric deflection, it probably makes sense to keep them as part of the PSF.  So the probable procedure will be to determine the mean center from a stack and then fix that for the individual PSF estimates, letting the PSF model itself include some centroid offset.

2. The `reflux` method is supposed to take an image of a star along with a model of the star at that location and just determine the flux and center value (if the latter is not being fixed).  The use case here is to compare models of the PSF to observed stars.  Either ones that were used to do the fit or possibly test stars that were reserved from the fit.  But to do that properly, you need to rescale it by the right flux.  Currently the implementation of this in Gaussian is lame -- it just uses the sum of the pixels in the array as the flux, which is wrong if there are any masked pixels.

3. Finally, Gary implemented a `chisq` method for PixelGrid which calculates the Taylor expansion of chisq with respect to the model parameters.  This allows you to defer the actual maximization until you have all the star measurements and can fit them all at once along with the interpolation.  I don't think this one is so important to implement for Kolmogorov, but it was important for PixelGrid when some of the stars have missing pixels.  There can be some parts of the model that are literally impossible to constrain from any single star image, so you kind of have to do this more sophisticated procedure.  So feel free to pass on this one, but if you feel ambitious and want to try to understand how that method is supposed to work from following through the PIxelGrid version, then feel free to include this here too.

And if you feel really ambitious, you could consider improving the Gaussian model to include these features as well.  :)

jmeyers314:
Sounds good.  `reflux` doesn't sound too tricky.

I'm not so sure about `chisq` though.  Is the optimization delay still possible if `chisq` isn't linear or quadratic in the parameters?

gbernstein:
[...]

PR #39, opened 2016-09-19
#38  Kolmogorov Model

jmeyers314:
This PR adds a sheared Kolmogorov PSF model.  The parameters are just a PSF size (fwhm), and shear components g1 and g2.  The center of the PSF is also optionally allowed to be a parameter.

I copied a number of unit tests from `test_pixel.py` to ensure that the above params are interpolatable across the field.

One aspect I know isn't implemented currently is marginalization over the flux and center (when center isn't part of the PSF).  The flux marginalization when the center is fixed might be analytic with a Gaussian prior --  I can think some more about that.  However, I don't think we can marginalize the flux or the center when the center is free without performing a numerical integral, which sounds expensive.  Probably for now we can get away without this, though, I'm hoping.


piff/kolmogorov_model.py
```diff
@@ -0,0 +1,311 @@
...
+            center = (cenu, cenv)
+        else:
+            params = np.array([ cenu, cenv, fwhm, g1, g2 ])
+            center = (0.0, 0.0)
+
+        # Also need to compute chisq
+        prof = self.getProfile(params) * flux
+        model_image = star.image.copy()
+        prof.shift(center).drawImage(model_image, method='no_pixel',
+                                     offset=(star.image_pos - model_image.trueCenter()))
```
cpadavis: I trust that you figured this out correctly, but since I get really quite confused about all this, I want to double check: in gaussian_model we compute the profile without the WCS, and compute the offset as

    galsim.PositionD(cenx, ceny) - star.image.trueCenter()

Here, you compute a shift in wcs coordinates (cenu = moment_center_u - image_pos_u) and then apply another offset in the drawing. This probably illustrates my ignorance of the galsim coordinate systems, but why not just group both into one thing, ie everything in prof.shift, or in the offset kwarg?

piff/kolmogorov_model.py
```diff
@@ -0,0 +1,311 @@
...
+            shape = galsim.Shear(g1 = -shape.g1, g2 = shape.g2)
+        # Next the rotation
+        shape = galsim.Shear(g = shape.g, beta = shape.beta + theta)
+        # Finally the shear
+        shape = shear + shape
+
+        # Another magic number below to convert flux -> actual flux for Kolmogorov.
+        flux = mom.moments_amp / 0.9053
+
+        center = image.wcs.toWorld(mom.moments_centroid) - image.wcs.toWorld(image_pos)
```
jmeyers314: Probably a lot of things here can be backported to the Gaussian module.  In fact, pretty much the only thing that's actually different conceptually between Gaussian and Kolmogorov are the magic numbers in this method.  Probably suggests we should use a common parent class.  Something like "GSObject_Model" might work.  I'll give that a think...

piff/kolmogorov_model.py
jmeyers314: It might just be in my head, but there, I found it helpful to separate the shift that was potentially part of the PSF model (cenu, cenv), from the shift that just indicates the subpixel offset of where we know (say, from Gaia), the star actually is w.r.t. the pixel grid.  I also thought it made sense to keep the PSF shift in world coords (since presumably that's how we'd want to interpolate it), and since the reference point shift (image_pos) was already documented as being in image coords I had to separate them.

FWIW, though, I am also easily confused by all the offsetting...

piff/kolmogorov_model.py
```diff
@@ -0,0 +1,311 @@
...
+        else:
+            params = np.array([ cenu, cenv, fwhm, g1, g2 ])
+            center = (0.0, 0.0)
+
+        # Also need to compute chisq
+        prof = self.getProfile(params) * flux
+        model_image = star.image.copy()
+        prof.shift(center).drawImage(model_image, method='no_pixel',
+                                     offset=(star.image_pos - model_image.trueCenter()))
+        chisq = np.sum(star.weight.array * (star.image.array - model_image.array)**2)
```
jmeyers314: I think the expression in `Gaussian` is probably just wrong.  The standard deviation erroneously subtracts the mean, and also takes a square root that I don't think we want.  There's also the factor of Npixels.

RobertLuptonTheGood: You have to be a bit careful with the 1/sigma^2 weighting as it biases the weighting away from the centre of for bright stars.  You want a good estimate everywhere, not just where you have nice stars.   You also care about the spatial distribution of the candidates -- is that being handled?

jmeyers314: >You have to be a bit careful with the 1/sigma^2 weighting as it biases the weighting away from the centre of for bright stars.

I'm not sure I understand.  Is it the contribution of the signal to the Poisson variance you're referring to?

>You also care about the spatial distribution of the candidates -- is that being handled?

Certainly not here.  Elsewhere in Piff maybe?

cpadavis: > You also care about the spatial distribution of the candidates -- is that being handled?

That should be handled by an interpolator type class. Piff would take the params [u, v], fwhm, g1, g2 from the kolmogorov model and then you could use your favorite interpolator to characterize the spatial distribution

piff/kolmogorov_model.py
rmjarvis: This being Python, I'd probably recommend having utility functions that exist outside of each of the classes, but which they can all use via duck typing as required.  I don't think there's likely to be much advantage to having a class hierarchy for this.

piff/kolmogorov_model.py
rmjarvis: The basic distinction, as Josh eludes to, is that `shift` is in world coordinates (aka field coordinates here in Piff), whereas `offset` is in image coordinates.  

Originally in GalSim we only had `shift`, but it quickly became apparent that sometimes it's more convenient to give the shift in pixel coordinates instead, which we put into the `drawImage` functionality as the parameter `offset`.

I think in this case, it is completely appropriate to use both translation options.  `shift(center)` moves the nominal center of the star in world coordinates, and then when you draw it, you provide an additional `offset` in pixel coordinates relative to the image center to account for the sup-pixel dithering of the particular observation.

piff/kolmogorov_model.py
rmjarvis: > You have to be a bit careful with the 1/sigma^2 weighting as it biases the weighting away from the centre of for bright stars. 

This is probably something that we'll want as an option, since we'll want to investigate what kind of weighting is best in practice.  We at least want the weights to be zero anywhere the weight image is zero, but I think it's probably an open question whether it is better to down-weight the noisier central pixels or not.  I've certainly heard this claim before, but I don't know if I've actually seen concrete evidence for it.  

Robert, have you investigated this effect on real profile fitting of real images in your SDSS work perhaps?  Or is this conclusion the result of thought experiments?

rmjarvis: > You also care about the spatial distribution of the candidates -- is that being handled?

The current interpolators don't do anything sophisticated with respect to the spatial distribution of the stars.  They just take all the stars with equal weight and do their fit.  My suspicion is that local interpolators like radial basis functions, kriging, etc. will be less sensitive to clumping of the stars than our current polynomial interpolators.  But that remains a project for future investigation.

RobertLuptonTheGood: I know the the 1/sigma^2 is a problem for photometry, and in SDSS I ended up using equal weights for all stars in the spatial interpolation (a KL-basis + polynomials) to avoid fitting bright stars in one corner of the field.  I have done spatial interpolation using the brightest star per cell, but didn't play extensively with weighting schemes (== I don't remember what/if I learned)

jmeyers314:
What's the PR review policy for Piff?  

I'd still like to deduplicate some of the Kolmogorov/Gaussian code (maybe add Moffat while I'm at it?), but is there anything else remaining for this?

cpadavis:
I think the PR review policy has been getting at least one person to look
over the code and give a thumbs up, and leave enough time for others to
comment as well. The code seems fine to me.

I think Mike is of the opinion that we should just duplicate the code in
the different places rather than create a middle GSObjectModel class. I bet
if we propose enough of these (there's Kolmogorov, Gaussian, Optics, and
potentially Moffat) we might be able to get him to relent :) Alternatively,
the GSObject might be a separate pull after we update the other bits.

rmjarvis:
> What's the PR review policy for Piff?

I'd like to go through it still.  Just haven't had a chance yet.

> I think Mike is of the opinion that we should just duplicate the code in
the different places rather than create a middle GSObjectModel class.

No.  I said that the common code should probably reside in helper functions, rather than a common base class.  Mostly because I foresee more cases where it will be useful to use some utility function than where the class would logically derive from this particular base.  So I think that would result in *less* duplication of code.

jmeyers314:
For deduplication, how about starting from a fiducial GSObject PSF:

```python
class GSObject_model:
    def __init__(gsobj, ...):
```

and then the PSF parameters become how much you need to shift, dilate, and shear that fiducial object to get a good match:

```python
    def getProfile(self, params):
        du, dv, dsize, dg1, dg2 = params
        return self.gsobj.dilate(dsize).shear(g1=dg1, g2=dg2).shift(du, dv)
```

  That automatically captures the important parts of Gaussian and Kolmogorov, I think, and is extensible to Moffats (with fixed beta and fixed truncation radius at least -- not sure if we care about that), or other profiles.  One thing it lacks is a quick and dirty fit method using HSM output like is currently in Gaussian and Kolmogorov.  (Although we could allow for such a function to be optionally passed into `__init__` if we wanted, or build it in with a class hierarchy.)

Thoughts?

jmeyers314:
I got the simplest possible unit test for `GSObjectModel` working now.  To generate Gaussian, Kolmogorov, or Moffat (with fixed beta) PSF models, you can do:

```python
fiducial_kolmogorov = galsim.Kolmogorov(half_light_radius=1.0)
fiducial_gaussian = galsim.Gaussian(half_light_radius=1.0)
fiducial_moffat = galsim.Moffat(half_light_radius=1.0, beta=3.0)
for fiducial in [fiducial_gaussian, fiducial_kolmogorov, fiducial_moffat]:
    model = piff.GSObjectModel(fiducial)
```

jmeyers314:
Think this is ready for review again.  

The one part I didn't tackle with the GSObjectModel design is how to implement this in the config layer, so ideas about that appreciated.

jmeyers314:
I think I figured out the config layer for GSObjectModel.  It relies on roundtripping eval/repr for the wrapped GSObject, which I think nearly always (always always?) works for GSObjects.

I did run into a different hiccup though in test_knn_interp.py.  Referring back [here](https://github.com/rmjarvis/Piff/issues/38#issuecomment-247805452)

>I think the general case would be:
>
>`center = image.wcs.toWorld(mom.moments_centroid) - image.wcs.toWorld(star.image_pos)`

Subtraction isn't defined when image.wcs is a `CelestialCoord`, so I think we still haven't actually discovered the general case.  Maybe there's a "locally affine" approximation that would work here?

tests/test_gsobject_model.py
```diff
@@ -0,0 +1,483 @@
...
+        print('True g1 = ',g1,', model g1 = ',fit.params[1])
+        print('True g2 = ',g2,', model g2 = ',fit.params[2])
+        print('True cenu = ', cenu, ', model cenu = ', fit.center[0])
+        print('True cenv = ', cenv, ', model cenv = ', fit.center[1])
+
+        # The sensitivity of this test is somewhat dependent on what the wcs is, and whether the fit
+        # center parameters are in image coords or world coords.
+        np.testing.assert_allclose(fit.params[0], size, rtol=1e-3)
+        np.testing.assert_allclose(fit.params[1], g1, atol=1e-3)
+        np.testing.assert_allclose(fit.params[2], g2, atol=1e-3)
```
rmjarvis: These tolerances for g1,g2 are less strict than with fastfit.  Was this intentional?  Or can we tighten them up?

tests/test_gsobject_model.py
```diff
@@ -0,0 +1,483 @@
...
+        mod = piff.GSObjectModel(fiducial)
+        star = mod.initialize(s)
+        print('Flux, ctr after reflux:',star.fit.flux,star.fit.center)
+        for i in range(3):
+            star = mod.fit(star)
+            star = mod.reflux(star)
+            print('Flux, ctr, chisq after fit {:d}:'.format(i),
+                  star.fit.flux, star.fit.center, star.fit.chisq)
+            # These fluxes are not at all close to influx.  Not sure why...
+            # np.testing.assert_almost_equal(star.fit.flux/influx, 1.0, decimal=2)
```
rmjarvis: This seems worth figuring out...

rmjarvis:
I see you've been working on this while I was reviewing.  I'm happy with the repr/eval trick, but I still think it's worth having direct Gaussian and Kolmogorov classes to hide this bit from the user.

rmjarvis:
> Subtraction isn't defined when image.wcs is a CelestialCoord, so I think we still haven't actually discovered the general case. Maybe there's a "locally affine" approximation that would work here?

You're right.  I think rather than `image.wcs` you probably need to use `image.wcs.local(star.image_pos)`.  I believe that should work for all kinds of WCS functions.

piff/util.py
```diff
@@ -140,3 +140,32 @@ def read_kwargs(fits, extname):
     assert len(data) == 1
     kwargs = dict([ (col, data[col][0]) for col in cols ])
     return kwargs
+
+def hsm(star):
+    """ Use HSM to measure moments of star image.
+    """
+    import galsim
+    image, weight, image_pos = star.data.getImage()
+    mom = image.FindAdaptiveMom(weight=weight, strict=False)
```
jmeyers314: Sure.  Done.

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+from .model import Model
+from .star import Star, StarFit, StarData
+
+
+class GSObjectModel(Model):
+    """ Model that takes a fiducial GalSim.GSObject and dilates, shifts, and shears it to get a
+    good match to stars.
+
+    :param gsobj:    GSObject to use as fiducial profile.
+    :param fastfit:  Use HSM moments for fitting.  Approximate, but fast.  [default: True]
```
jmeyers314: Yes, best not to lie in the docstring...  Fixed.

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+        fiducial_star = Star(sd, None)
+        flux, centroid, size, shape, flag = self.hsm_moments(fiducial_star)
+        if flag != 0:
+            raise RuntimeError("Error calculating fiducial moments for this gsobject.")
+        self._fiducial_size = size
+        self._fiducial_shape = shape
+        self._fiducial_flux = flux
+        self._fiducial_centroid = centroid
+
+    def hsm_moments(self, star):
```
jmeyers314: Sure.  Eliminated this method and using the utils version instead.

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+
+        :returns: `chi` as a flattened numpy array.
+        """
+        image, weight, image_pos = star.data.getImage()
+        flux, cenu, cenv, size, g1, g2 = lmparams.valuesdict().values()
+        # Fit cenu and cenv regardless of force_model_center.  The difference is whether the fit
+        # value is recorded (force_model_center=False) or discarded (force_model_center=True).
+        prof = self.gsobj.dilate(size).shear(g1=g1, g2=g2).shift(cenu, cenv) * flux
+        model_image = image.copy()
+        prof.drawImage(model_image, method='no_pixel',
```
jmeyers314: Sure.  Most of the tests only make sense for `include_pixel=False`, but I think that's okay.  To make sure we're covered, I added a new test for when `include_pixel=True`.

tests/test_gsobject_model.py
jmeyers314: Just kidding.  These pass at 14 digits of accuracy, not 2.

I went through and upped the test requirements throughout this file.

tests/test_gsobject_model.py
jmeyers314: I was able to bump these to atol=1e-6, which is slightly less good than with `fastfit` which I bumped to 1e-7.  Kind of cool that HSM is more accurate than model fitting even when there's no noise and it's the correct model.  Maybe could have been more aggressive with the lmfit stopping criteria...  not that it matters at 1e-6 ...

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+        flag = 0 if results.success else 1
+
+        return dflux, dcenu, dcenv, dsize, dg1, dg2, flag
+
+    def fit(self, star, fastfit=False, logger=None):
+        """Fit the image either using HSM or lmfit.
+
+        If `fastfit` is True, then the galsim.hsm module will be used to estimate the parameters of
+        the Kolmogorov PSF from image moments.  If `fastfit` is False, then the Levenberg-Marquardt
+        minimization algorithm as implemented by lmfit will be used instead.  The latter should
```
jmeyers314: Sure.  Removed.

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+        generally be more accurate, but slower due to the need to iteratively propose model
+        improvements.
+
+        :param star:    A Star to fit.
+        :param fastfit: Use fast HSM moments to fit? [default: False]
```
jmeyers314: Yep.  Fixed.

piff/gsobject_model.py
```diff
@@ -0,0 +1,341 @@
...
+                                           flux = star.flux*flux_ratio,
+                                           center = star.fit.center,
+                                           chisq = new_chisq,
+                                           dof = np.count_nonzero(weight.array) - 1,
+                                           alpha = star.fit.alpha,
+                                           beta = star.fit.beta))
+
+
+# Wrapped Gaussian model for backwards compatibility.
+# Should we do this for Kolmogorov and Moffat too?
```
jmeyers314: I now have named versions as well as the eval/repr trick, and tests for all.  :)

tests/test_simple.py
```diff
@@ -240,7 +240,7 @@ def test_single_image():
     # Round trip to a file
     psf.write(psf_file, logger)
     psf = piff.read(psf_file, logger)
-    assert type(psf.model) is piff.Gaussian
+    assert type(psf.model) is piff.gsobject_model.Gaussian
```
jmeyers314: Yes.  Fixed.

tests/test_simple.py
```diff
@@ -334,7 +334,6 @@ def test_single_image():
     p.communicate()
 
 if __name__ == '__main__':
-    test_Gaussian()
-    test_Mean()
+    # test_Gaussian()
+    # test_Mean()
```
jmeyers314: Fixed.

jmeyers314:
I think I've hit everything @rmjarvis .  Let me know if you have any more comments, especially concerning chisq/alpha/beta.

piff/gsobject_model.py
```diff
@@ -0,0 +1,389 @@
...
+        centroid = galsim.PositionD(cenu, cenv)
+        shape = galsim.Shear(g1=g1, g2=g2)
+
+        if flag:
+            raise RuntimeError("Failed measuring HSM moments of star.")
+
+        dflux = flux/self._fiducial_flux
+        dcentroid = centroid - self._fiducial_centroid
+        dsize = size/self._fiducial_size
+        dshape = shape - self._fiducial_shape
```
rmjarvis: I think all this fiducial stuff isn't working the way you want.  I was investigating why the tests in test_single_image (in test_simple.py) lost 1 d.p.  You use fastfit there, which should be equivalent to the old code, so the agreement shouldn't have changed.

I think I tracked it down to this bit of code.  Here you return the values relative to the fiducial values, but AFAICT you don't ever re-apply the fiducual values to get the real size, g1, g2, etc.  Furthermore, it looks like just a slightly lower precision because the fiducial flux, size, g1,g2 are in this case: 1.0000512599945068, 1.0000552542315129, -7.9303995699720791e-17, -9.589749686788564e-18.  So the deltas are not much different from the true values.

So I think there are probably two bugs here.  First, it seems like there is something wrong with how you are computing the fiducial values if this is what comes back for them.  And second, there are presumably places where you need to be adding or multiplying them back in to get the values you actually care about to send to the interpolator.

Or possibly, you may just want to scrap the fiducial stuff rather than fix it, since it's apparently not doing anything right now other than mess up the 5th decimal place.  Which means the code would probably work fine (or better) without it.

piff/gsobject_model.py
```diff
@@ -0,0 +1,389 @@
...
+        """Generate an lmfit.Parameters() instance from arguments.
+
+        :param star:         A Star from which to initialize parameter values.
+        :param vary_params:  Allow non-flux and non-center params to vary?
+        :param vary_flux:    Allow flux to vary?
+        :param vary_center:  Allow center to vary?
+
+        :returns: lmfit.Parameters() instance.
+        """
+        import lmfit
```
rmjarvis: BTW, you also need to add lmfit to the dependency list in setup.py.

jmeyers314: Sure.  Treecorr I think too, right?  I noticed one of the tests required this and I hadn't pip installed it yet.

rmjarvis: Ah, yes.  Probably so.

We could consider making dependencies optional at some point if our dependency burden becomes too significant, but for now let's try to make sure we explicitly include everything that is needed for the unit tests to run properly.

rmjarvis: (lmfit is a very light dependency btw, so no problem with that one!)

piff/gsobject_model.py
jmeyers314: For Gaussian with sigma=1.0, the default drawImage scale is ~0.85.  Running HSM on the resulting image gives the flux, size, g1, and g2 params you list above.

If I force the scale to 0.2, then the HSM flux and size come out exactly 1.0.  The g1 and g2 are still a bit off from zero, but I'm sure we don't care about 10^-17 (although it is a bit weird, since the image is perfectly symmetric).

If I force the scale to 0.2 inside `GSObjectModel.__init__`, then all the unit tests pass with their old tolerances.  So I think the different wcses is the issue.  Only question then is how to address it.  Ironically, my first instinct would have been to use the `wmult` keyword just removed from GalSim.  In lieu of that, maybe bumping some of the accuracy parameters in gsparams would work instead?

rmjarvis: I see.  I think maybe I have a little better understanding of how this fiducial stuff works now.  (I was confused earlier about what it was intended for -- mostly because I only looked at the parameters for the Gaussian case, where things are all supposed to be either 1 or 0.  And for Gaussian, it's only close but not perfect because of the GalSim automatic pixel scale bit.

So, I think we probably do want to bump up the GalSim accuracy slightly then.  Probably maxk_threshold=folding_threshold=1.e-4 would be appropriate.  And I don't think that would slow things down all that much for these nicely smooth profiles.  In particular, maxk_threshold is the one that mostly governs the pixel scale in real space.

rmjarvis: I had another idea about this on the drive home.  I think probably what should belong to the star is the hsm measurement of the data.  Then I think this is a reasonable flow for repeated calls to `fit`:
1. First pass checks for the stored hsm measurement, sees that it isn't there, and calculates it.  It compares to the fiducial stuff in the class to derive estimates of size, flux, etc.  i.e. the same thing that is currently being done.
2+. Later passes see that there is already a stored hsm result, which means that it should also have a `fit.params` with the current estimate of the size, shape, etc.  So now, rather than run hsm of the data, it draws the current model using the same image size, scale, etc. and runs hsm on that.  Then it can refine the estimate of the params by comparing the two hsm results.

A couple thoughts about this proposed process.  First I'm not sure where to store the hsm results in the star object.  I guess star.data.properties, but I haven't thought through whether there are any consequences to that.  I think it should be fine though.

Second, I think a few iterations of this may be as accurate as the lmfit stuff.  If so, since we will typically run through a few passes anyway to deal with outliers, it's possible this could just become the one implementation of fit in this class, rather than use the lmfit stuff.  (OTOH, it's also possible that it won't.  Just it's worth checking to see.)

jmeyers314: I reworked things so that there's no longer any `GSObjectModel._fiducial_params`.  Instead, all of the moment comparisons are between the moments of the model given `fit.params` and the stored data moments in `data.properties['hsm']`.  That also means I never have to draw `self.gsobj` without knowing what `wcs` to use.  So, a nice improvement -- thanks.

I kept the `lmfit` option for now, since at the very least it represents a different kind of "loss function" and also seemed to perform better on some of the unit tests.  

That said, the unit tests seem pretty delicate -- some are probabilistic, there's at least one race condition, and it seemed like some tests would transition from passing to failing if more iterations of fastfit were used...  So I made sure everything passed three times before sending this :).

rmjarvis:
Thanks Josh.  I'm happy with this change to finding differential steps for the flux, shape, etc.  I think this is more in line with the way we've done other models (PixelGrid in particular) than the way Gaussian used to work, so hopefully that will be a paradigm we can stick to going forward.  Merging now.

PR #42, opened 2016-10-28
Gaussian Process Interpolation

jmeyers314:
Added a GP Interpolant.  I eventually settled on using the latest version of scikit-learn for the GP framework.  It has nice features like being able to compose kernels from other kernels.  

There were a few features missing, however, such as the ability to assert a kernel directly measured from the covariance of the data, or use 2D kernels with fully general anisotropic covariance.  So I added these as `EmpiricalKernel` and `AnisotropicRBF`. (RBF = radial basis function)

I think the only part left out at the moment is the ability to write an `EmpiricalKernel` to a fits extension, which is hard because `EmpiricalKernel` takes a generic python callable as an argument.  We could potentially make a `LookupTable2DKernel` if desired, though, which may accomplish much the same thing and still be serializable.

Finally, note that I merged in #38 here already, so the diffs here will reflect both sets of changes until that one gets merged to master.


rmjarvis:
Hm.  This works on my laptop, but one of your new tests is failing on Travis.  Only 0.09% failure rate though, so maybe this is just a numerical rounding difference between the two systems.

jmeyers314:
I got it to pass by changing the test requirements a bit.  I really don't know how well we expect the interpolator to work, so maybe this is okay?

I did notice one other peculiarity though.  I've largely switched to the py.test test runner (mostly just to get my dots back in python3, but also for the nice colorful output), and with that, I see a failure in test_simple.py.  When I run nosetests instead the output report is "Ran 2 tests ..." (no failures) even though there are 3 tests in the file.  When I run it directly (python test_simple.py), I get a segfault.  So that makes me suspicious of the travis checks too.

rmjarvis:
Now it's flipped.  On Travis is passes, but my laptop is failing:
```
============================================= FAILURES =============================================
___________________________________________ test_grf_psf ___________________________________________

    def test_grf_psf():
        rng = galsim.BaseDeviate(987654334587656)
        ntrain, nvalidate, nvisualize = 100, 1, 21
        training_data, validation_data, visualization_data = \
            make_grf_psf_params(ntrain, nvalidate, nvisualize)
        kernel = 1*kernels.RBF(0.3, (1e-1, 1e1))
        # We probably aren't measuring fwhm, g1, g2, etc. to better than 1e-5...
        kernel += kernels.WhiteKernel(1e-5, (1e-7, 1e-1))
        for npca in [0, 5]:
            for optimizer in [None, 'fmin_l_bfgs_b']:
                check_gp(training_data, validation_data, visualization_data, kernel,
>                        npca=npca, optimizer=optimizer, filename="test_gp_grf.fits", rng=rng)

test_gp_interp.py:532: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test_gp_interp.py:469: in check_gp
    validate(validate_stars, interp)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

validate_stars = [<piff.star.Star object at 0x117e47110>]
interp = <piff.gp_interp.GPInterp object at 0x11b1bd950>

    def validate(validate_stars, interp):
        """ Check that global PSF model sufficiently interpolates some stars.
        """
        # Noiseless copy of the PSF at the center of the FOV
        for s0 in validate_stars:
            s0 = mod.initialize(s0)
            s0 = mod.fit(s0)
            s0 = mod.reflux(s0)
    
            s1 = interp.interpolate(s0)
            s1 = mod.reflux(s1)
            print()
            print(("s0: "+"{:< 8.4f} "*len(s0.fit.params)).format(*s0.fit.params))
            print(("s1: "+"{:< 8.4f} "*len(s1.fit.params)).format(*s1.fit.params))
            print("s0 flux:", s0.fit.flux)
            print("s1 flux:", s1.fit.flux)
            print()
            print('Flux, ctr, chisq after interpolation: \n', s1.fit.flux, s1.fit.center, s1.fit.chisq)
>           np.testing.assert_allclose(s1.fit.flux, s0.fit.flux, rtol=3e-3)
E           AssertionError: 
E           Not equal to tolerance rtol=0.003, atol=0
E           
E           (mismatch 100.0%)
E            x: array(0.9967843740000385)
E            y: array(1.0000000000000007)

test_gp_interp.py:437: AssertionError
--------------------------------------- Captured stdout call ---------------------------------------


Initial state
-------------
chisq: 101286.185488    dof: 102300
chisq/dof: 0.990089789716

iteration: 0  chisq: 101476.490702  dof: 102300
iteration: 1  chisq: 101476.490851  dof: 102300
1**2 * RBF(length_scale=0.3) + WhiteKernel(noise_level=1e-05)

s0:  0.1632  -0.3737   0.6368   0.0212  -0.0026  
s1:  0.1641  -0.3723   0.6348   0.0179   0.0002  
s0 flux: 1.0
s1 flux: 0.996784374

Flux, ctr, chisq after interpolation: 
 0.996784374 (0.0, 0.0) 3.2767858177e-05
```
Again, it's just barely missing, so probably the test just needs to be tuned differently.

Also, I'm not seeing any failure in test_simple using either py.test or nosetests on my laptop.  I'll try out some other systems to see if any of them give a seg fault, but my first guess is that maybe it's from the piffify call in test_single_image, which might not show up properly as an error that nosetests can catch.  So you could try running `piffify simple.yaml` to see if that's where the error is happening.

rmjarvis:
On another system, I'm getting errors that look like
```
_____________________________ ERROR collecting tests/test_gp_interp.py _____________________________
test_gp_interp.py:19: in <module>
    import piff
/astro/u/mjarvis/lib/python2.7/site-packages/Piff-0.1-py2.7.egg/piff/__init__.py:86: in <module>
    from .gp_interp import GPInterp, EmpiricalKernel, AnisotropicRBF
/astro/u/mjarvis/lib/python2.7/site-packages/Piff-0.1-py2.7.egg/piff/gp_interp.py:187: in <module>
    from sklearn.gaussian_process.kernels import StationaryKernelMixin, NormalizedKernelMixin, Kernel
E   ImportError: No module named kernels
```
which I think means the sklearn dependency in setup.py needs to have a >= version number to set the minimum version that has this feature.

rmjarvis:
On that same machine, once I installed sklearn 0.18.1, I did get an error on test_simple.py that crashed the test with
```
 : cannot connect to X server localhost:11.0
```
which I think is related to matplotlib.  Not sure if that's the same thing that is happening on your system or not, since it happens for both py.test and nosetests.  I'll see if I can figure out a workaround for the case where there is no X server available.

rmjarvis:
Josh, I'm pretty sure I know what the seg fault is from.  The latest pip version of fitsio has a bug reading string items where it callocs one byte too few for the sting.  @esheldon [fixed](https://github.com/esheldon/fitsio/commit/25abfede2a67b51dd4418b000d48809068eb1ea9) this on master, so if you git clone fitsio and install it manually, it should work.  That's what I currently do on travis as a stopgap until Erin releases the next version of fitsio with the bug fix.  I just forgot when you mentioned that you were getting seg faults that this was probably the problem.

jmeyers314:
Hmmm.... `piffify simple.yaml` appears to run fine.  (At least, it doesn't crash.  But it doesn't print anything to stdout either).

I tried upgrading fitsio, but I'm still getting the segfault.  I narrowed it down to 

test_simple.py:280  (calls stats.compute(psf, orig_stars)
stats.py:351 (calls self.measureShapes(psf, stars, logger))
stats.py:147 (calls piff.util.hsm(star))
util.py:172 (calls image.FindAdaptiveMom)

If I put an ipdb breakpoint right before the FindAdaptiveMom and try to plot the image, I also get a segfault, so I think something may be weird with the image.

rmjarvis:
> I tried upgrading fitsio, but I'm still getting the segfault.

I haven't managed to find a system that reproduces your behavior.  But given that it's in the stats module, it's possible that it's related to matplotlib.pyplot.  Do you also get segfaults in the `no_pyplot` branch?

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def initialize(self, stars, logger=None):
+        """Initialize both the interpolator to some state prefatory to any solve iterations and
+        initialize the stars for use with this interpolator.
+
+        :param stars:   A list of Star instances to interpolate between
+        :param logger:  A logger object for logging debug info. [default: None]
+        """
+        self.solve(stars, logger=logger)
+        return self.interpolateList(stars)
```
rmjarvis: initialize isn't supposed to do any actual solving.  It's really just supposed to do any necessary setup of either the `star.fit` objects or the interpolator itself (e.g. setup numpy arrays to be the right size given which parameters are being interpolated). 

I think in this case, there is nothing in the GPInter object to set up.  This might even be ok to just `return stars`.  But if you want, you could probably set up the `star.fit` attributes to be not `None`.  (I admit I'm still a little hazy on why Gary needed the initialize method for BasisPolynomial.)

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def solve(self, stars=None, logger=None):
+        """Set up the GaussianProcessRegressor to be able to predict.
```
rmjarvis: GaussianProcessRegressor is an implementation detail.  Since `solve` is a public API method, this should be reworded.  Probably just replace with `this GPInterp object`.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+        data['INIT_THETA'] = init_theta
+        data['FIT_THETA'] = fit_theta
+        data['X'] = self._X
+        data['Y'] = self._y
+
+        fits.write_table(data, extname=extname+'_kernel')
+
+    def _finish_read(self, fits, extname):
+        data = fits[extname+'_kernel'].read()
+        optimizer = self.gp.optimizer
```
rmjarvis: redundant?

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+import numpy as np
+
+class GPInterp(Interp):
+    """
+    An interpolator that uses sklearn.gaussian_process to interpolate a single surface.
+
+    :param  attr_interp: A list of star attributes to interpolate from
+    :param  kernel:      A Kernel object indicating how the different stars are correlated.
+    :param  optimizer:   Optimizer to use for optimizing the kernel.  Set to `None` to skip
+                         kernel optimization.
```
rmjarvis: Explain what the default is.  Also, can we come up with a better name than 'fmin_l_bfgs_b'?  I guess that's sklearn's name for this, but it's highly non-intuitive, so if we can give a more legible alias, I think that would be helpful.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+
+from sklearn.gaussian_process.kernels import StationaryKernelMixin, NormalizedKernelMixin, Kernel
+from sklearn.gaussian_process.kernels import Hyperparameter
+
+class EmpiricalKernel(StationaryKernelMixin, NormalizedKernelMixin, Kernel):
+    """ A GaussianProcessRegressor Kernel that just wraps an arbitrary python function.
+
+    :param  fn:  Python callable that accepts two numpy arrays indicating the differences in two
+                 covariates and returns a numpy array indicating the covariances implied by those
+                 differences.
```
rmjarvis: I'm not sure how this class will be useful if it's not serializable.  Typical Piff usage will be via config files that will produce output files with the PSF information.  Then these files will be read in to generate PSF images.  So if we can't provide the input as a string and can't serialize it into the output file, then I don't think this is going to be usable in practice.

I think both problems can fixed by having the input be an eval-ed string that evaluates as a lambda function, rather than have it already be a Python function.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+        self.gp.optimizer = optimizer
+        # Now that gp is setup, we can restore it's initial kernel.
+        self.gp.kernel.theta = self.kernel.theta = data['INIT_THETA'][0]
+
+
+
+from sklearn.gaussian_process.kernels import StationaryKernelMixin, NormalizedKernelMixin, Kernel
+from sklearn.gaussian_process.kernels import Hyperparameter
+
+class EmpiricalKernel(StationaryKernelMixin, NormalizedKernelMixin, Kernel):
```
rmjarvis: Why is this called `Empirical`?  That implies to me something based on the data.  Not a kernel with a given explicit function.  It seems like `FunctionKernel` or `ExplicitKernel` would be more apt.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+          ...         ...            ...         ...    ...               ...         ]
+          th[]        th[]           th[]        ...    exp(th[n-2])      0           ]
+          th[]        th[]           th[]        ...    th[n*(n+1)/2-1]   exp(th[n-1])]]
+
+    I.e., the inverse covariance matrix is Cholesky-decomposed, exp(theta[0:n]) lie
+    on the diagonal of the Cholesky matrix, and theta[n:n*(n+1)/2] lie in the lower triangular
+    part of the Cholesky matrix.  This parameterization invertably maps all valid n x n
+    covariance matrices to R^(n*(n+1)/2).  I.e., the range of each theta[i] is -inf ... inf.
+
+    :param  invLam:  Inverse covariance matrix of radial basis function.
```
rmjarvis: This one is also problematic.  How is the user supposed to specify this in the config file?  We need to have Piff create the invLam function for them based on some measurement of the data.  Preferably with something like `kernel='anisotropic'` as the argument to `GPInterp`.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+
+                dX = X[:, np.newaxis, :] - X[np.newaxis, :, :]
+                dist_grad = np.einsum("ijk,lkm,ijm->ijl", dX, invLam_grad, dX)
+                K_gradient = -0.5 * K[:, :, np.newaxis] * dist_grad
+                return K, K_gradient
+        else:
+            return K
+
+    @property
+    def hyperparameter_cho_factor(self):
```
rmjarvis: cho looks to me like a typo of chi.  Recommend spelling out cholesky.

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+from .interp import Interp
+from .star import Star, StarFit
+
+import numpy as np
+
+class GPInterp(Interp):
+    """
+    An interpolator that uses sklearn.gaussian_process to interpolate a single surface.
+
+    :param  attr_interp: A list of star attributes to interpolate from
```
cpadavis: we should probably also change knn_interp.py and decam_wavefront.py as well, since they also use this attr_interp style

piff/gp_interp.py
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+        self.gp = GaussianProcessRegressor(self.kernel, optimizer=optimizer)
+
+    @staticmethod
+    def _eval_kernel(kernel):
+        from sklearn.gaussian_process.kernels import WhiteKernel, RBF, Matern, RationalQuadratic
+        from sklearn.gaussian_process.kernels import ExpSineSquared, DotProduct, PairwiseKernel
+        from numpy import array
+        return eval(kernel)
```
cpadavis: Should there be some sort of error catching to say what valid kernels are available? I can imagine it would be really frustrating if something types 'ExpSinSquared' and doesn't understand why it's failing...

piff/interp.py
```diff
@@ -173,7 +173,7 @@ def _finish_write(self, fits, extname):
         :param fits:        An open fitsio.FITS object
         :param extname:     The base name of the extension
         """
-        pass
+        raise NotImplementedError("Derived classes must define the _finish_write method.")
```
cpadavis: iirc we had figured _finish_write might be a method that wasn't needed by all classes, and so if it wasn't needed, then it just wouldn't do anything...

piff/interp.py
```diff
@@ -216,4 +216,4 @@ def _finish_read(self, fits, extname):
         :param fits:        An open fitsio.FITS object.
         :param extname:     The base name of the extension.
         """
-        pass
+        raise NotImplementedError("Derived classes must define the _finish_read method.")
```
cpadavis: see above comment on _finish_write

piff/interp.py
rmjarvis: I think that was the rationale.  But given that even `MeanInterp` implements this, I suspect all interpolant classes will need to do something here.  

And if not, it's probably a good idea to explicitly override this method and just have it `pass` with a comment saying why nothing else is required.

piff/gp_interp.py
rmjarvis: I think I prefer the name `keys` for this, but if people prefer `attr_interp` that's ok too.  I mostly just want the same name for the same concept across different classes.

piff/gp_interp.py
cpadavis: Does this mean that for every interp, we need to be able to catch if the solving has not occurred? It seems weird to me to have a silent failure for interpolating with an interpolant that hasn't solved...

rmjarvis: I don't think I understand the question.  My suggestion was to just `return stars`.  Not continue to run `return self.interpolateList(stars)`.  That was an implementation detail of Gary's BasisPolynomial interpolator.  It installed a nummy solution that was just a constant and then interpolated with that to get some things to have the right structures.  I don't think anything like that is required here, so the `initialize` function can probably be essentially a no op.

jmeyers314: I'm trying out making `kNNInterp.initialize` just `return stars`.  That breaks `test_knn_interp.py:test_yaml()` though, due to a RuntimeError in `GSObjectModel.moment_fit` caused by a failure in `galsim.hsm.FindAdaptiveMom`.  Only 1 of the 117 stars in the test has this problem (it's one with a big weight=0 gash through the middle of the image, so not terribly surprising).

Is there a piffonic way to handle failed model fits?  Presumably somewhere near the offending line in simplepsf.py:138:

```python
                self.stars = [self.model.fit(s) for s in self.stars]
```

jmeyers314: Looks like there's handling of failed `reflux`ing a little further down the file.  I guess I can just copy that...

rmjarvis: Feel free to change what the "piffonic way" is here.  I think we're all still trying to feel out what that should be, so if something different makes sense, go with it.

piff/gp_interp.py
jmeyers314: Done.  `attr_interp` -> `keys`  and `attr_target` eliminated.

piff/gp_interp.py
jmeyers314: Done. `GPInterp.initialize()`  (and also `kNNInterp.initialize()`) simply return `stars`.

piff/gp_interp.py
jmeyers314: >How is the user supposed to specify this in the config file? We need to have Piff create the invLam function for them based on some measurement of the data.

My thought on this one was that the user would just supply an initial `invLam`, and then `GPInterp.fit` would optimize this (if `optimizer != None`) based on the data.  (Can the config yaml handle a 2x2 numpy array, or list of lists, as a parameter?) There's nothing in this PR currently for generating a good initial guess, though that's certainly a reasonable idea.

piff/gp_interp.py
jmeyers314: Yep.  Thanks.

piff/gp_interp.py
jmeyers314: Changed name to `ExplicitKernel` and changed input arg to be eval-able lambda string.

piff/gp_interp.py
jmeyers314: Sure. Done.

piff/gp_interp.py
jmeyers314: I went ahead and switched this to just a boolean: `optimize`.

jmeyers314:
Thanks @rmjarvis and @cpadavis .  I think I've addressed everything now.

rmjarvis:
I'm getting a unit test failure on my laptop.  Seems not to happen on Travis apparently.
```
======================================================================
FAIL: test_gp_interp.test_guess
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Mike/rmjarvis/Piff/tests/test_gp_interp.py", line 688, in test_guess
    np.testing.assert_allclose(inferred_scale_length, 0.3, rtol=0.1)
  File "/Users/Mike/lib/python2.7/site-packages/numpy/testing/utils.py", line 1392, in assert_allclose
    verbose=verbose, header=header)
  File "/Users/Mike/lib/python2.7/site-packages/numpy/testing/utils.py", line 739, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=0.1, atol=0

(mismatch 80.0%)
 x: array([ 0.334331,  0.32425 ,  0.339469,  0.334682,  0.33337 ])
 y: array(0.3)

----------------------------------------------------------------------
```
It seems suspiciously close to 0.333, but that could be a coincidence.

Also, the unit tests are pretty slow.  135 seconds for test_gp_interp compared with 151 for everything else put together.  I haven't been targeting 1 seconds each like I try to for GalSim (since the nature of the tests is rather different, so sometimes we need several seconds to do something reasonable), but this seems like too long.  

Can we try to use the `if __name__ == '__main__'` trick to make faster tests via nosetests, but leave the longer more precise tests when running from the command line?  Also might be worth profiling to see if there is something obvious that is causing a big slow down that we might be able to speed up.  (I didn't try any profiling myself yet.)

jmeyers314:
>It seems suspiciously close to 0.333, but that could be a coincidence.

I'm pretty sure that's a coincidence.  The test is trying to find 0.3 using 5 different initial guesses, but each time with the same simulated data.  So it's likely that this particular pseudo-random instantiation of the data has a maximum likelihood around 0.333, and each initial guess is finding that.

I went ahead and just bumped the tolerance up a bit.

>Can we try to use the if __name__ == '__main__' trick to make faster tests via nosetests ... ?

Done.  Now it looks like the gp stuff is ~70 seconds on top of ~210 seconds for everything else (on travis).  And no coverage lost.

I did need to bump another tolerance up -- from 3e-3 to 1e-2 on the interpolated and refluxed flux of a test star.  Unlucky rng maybe?  (or maybe previously we were getting a lucky rng?)  I'd kind of like this to be working to better than 1% there, but I don't have a good numerical argument for how well it should be working, so maybe 1% is okay.

jmeyers314:
... and setting `fastfit=True` brings it down to ~30 seconds.

PR #43, opened 2016-11-08
Python 3 compatibility for polynomial test

gogrean:
This is a set of changes meant to make the polynomial test run in Python 3. It mostly consists of several changes of xrange to range, and many messy bytes-to-str and str-to-bytes conversions required for reading/writing the FITS files with `fitsio`. I have not tested the code in Python 2, and it's likely some of the string conversions only work in Python 3. If that's the case, I could add some checks for the Python version used, or perhaps you prefer to make a Python 3 branch instead. 

Some parts of the code that I modified would be a bit more neat if using `six`. I have added this as a TODO.


piff/util.py
```diff
@@ -62,6 +62,10 @@ def make_dt_tuple(key, t, size):
         # Note: this works for either arrays or strings
         size = len(value)
         t = type(value[0])
```
gogrean: Already addressed in the else clause.

gogrean:
Closing in favor of #44

rmjarvis:
Sorry I haven't looked at the other one yet.  I'll take a look at this this week.  Did you check if this still works in Python 2.7?

I should really get Travis set up here to automatically test this for you, including both Python 2 and Python 3, but I haven't done that yet...

rmjarvis:
(Oops.  I meant for that comment to be on #44...)

PR #44, opened 2016-11-17
Py3 compatibility for KNN interpolation test

gogrean:
Pickling the test data in Python 3 produces null bytes (e.g. "\x00\x00" in Python 3 vs "\\\x00\\\x00" in Python 2) that later cause the input to be truncated. Fixed by encoding the pickled stuff in base64. Files written with Python 2 can still be read with Python 3, and the other way around, at least as far as the test files are concerned.

Also fixed TypeError triggered by a list-integer comparison in `des/decaminfo.py` by first converting `chipnums` to a numpy array.

This also includes a set of changes meant to make the polynomial test run in Python 3. It mostly consists of several changes of xrange to range, and many messy bytes-to-str and str-to-bytes conversions required for reading/writing the FITS files with fitsio.

Some parts of the code that I modified would be a bit more neat if using six. I have added this as a TODO.


rmjarvis:
Sorry I haven't looked at the other one yet. I'll take a look at this this week. Did you check if this still works in Python 2.7?

I should really get Travis set up here to automatically test this for you, including both Python 2 and Python 3, but I haven't done that yet...

gogrean:
I now have Piff working with Python 2.7 as well, so I can test both. Here's what I have so far:

tests that pass both in 2.7 and 3.4:

- test_star.py
- test_helper.py
- test_pixel.py
- test_twodstats.py
- test_simple.py
- test_poly.py
- test_knn_interp.py

tests that only pass in Python 3:

- test_optics.py (working on it...)

gogrean:
Closing in favor of #45

PR #45, opened 2016-11-19
Tests pass in Python 2.7 and 3.4

gogrean:
I believe all tests should now run both in Python 2.7 and 3.4 (given the changes made, it's likely 3.x not just 3.4). I'm going to close the previous PR and just leave this one up. I copied below the text from #44.

---

Pickling the test data in Python 3 produces null bytes (e.g. "\x00\x00" in Python 3 vs "\\x00\\x00" in Python 2) that later cause the input to be truncated. Fixed by encoding the pickled stuff in base64. Files written with Python 2 can still be read with Python 3, and the other way around, at least as far as the test files are concerned.

Also fixed TypeError triggered by a list-integer comparison in des/decaminfo.py by first converting chipnums to a numpy array.

This also includes a set of changes meant to make the polynomial test run in Python 3. It mostly consists of several changes of xrange to range, and many messy bytes-to-str and str-to-bytes conversions required for reading/writing the FITS files with fitsio.

Some parts of the code that I modified would be a bit more neat if using six. I have added this as a TODO.


rmjarvis:
Thanks Georgiana.  I ended up merging this manually, since I messed up the Github automatic merge by merging in #38 before this.  I also changed the way the encode/decode stuff is done a bit, which I think is a bit cleaner.  cf. 0bd8c48

gogrean:
Looks good, thanks. Are we trying to catch more than a possible AttributeError on lines 155 and 160 in utils.py? That is the only exception that I have been able to generate, though I have only tried it in Python 3.

PR #46, opened 2016-11-19
Fixed poly test when run on multiple cores

### LINKED ISSUE #41: `nosetests test_poly` isn't thread safe
A few times every ten attempts or so, I'm getting errors when running `nosetests test_poly.py` from the test directory.  Looks like `test_poly_load_save` and `test_poly_load_err` are both calling `poly_load_save_sub` on different threads, leading to problems.

gogrean:
Fixes #41.

### DIFF
--- a/tests/test_poly.py
+++ b/tests/test_poly.py
@@ -289,8 +289,8 @@ def test_poly_guess():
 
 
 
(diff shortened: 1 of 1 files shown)


rmjarvis:
Thanks for this too.  :)

PR #48, opened 2016-12-05
Don't use pyplot or numpy.random state machines

rmjarvis:
I was having trouble getting the unit tests to run on a machine that couldn't access the X server.  I kept getting:
```
 : cannot connect to X server localhost:11.0
```
which bombed out like a seg fault, not just raising a python Exception.  After messing with different iterations of the `plt.use('Agg')` function for a while, I eventually realized that we shouldn't be changing the user's pyplot state.  This is actually pretty bad form for a library.  Much better etiquette to use the object oriented API of matplotlib instead.

Anyway, when I made the switch to never importing `matplotlib.pyplot`, the errors finally stopped, so I think this is the way to go.  For now, there is a bit of extra boiler plate, since you have to build up the subplots manually, but once matplotlib 2.0 comes out, the API will be almost equivalent to the current `pyplot`-based API.

Also, while I was at it, I switched up our unit tests to not use the `numpy.random` global state either.  It's not as important, since we only do that in the tests, but it might lead to issues when running nosetests in multiprocessing mode.  So I switched to using `numpy.random.RandomState` instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,8 @@ cache:
     - $HOME/GalSim-1.4.2
     - $HOME/tmv-0.73
     - $HOME/boost_1_61_0
-    - $HOME/.cache/matplotlib
+    - $HOME/.cache
+    - $HOME/.matplotlib
 
 install:
     - pip install -r requirements.txt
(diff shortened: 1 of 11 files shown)


rmjarvis:
Thanks Chris.  I'll plan on merging this Monday, in case anyone else would like to take a look.

PR #53, opened 2017-06-21
Add ability to specify a WCS other than what is in the image file

rmjarvis:
... and other stuff.  Also cf. galsim_extra PR [here](https://github.com/esheldon/galsim_extra/pull/8/files), especially the wrongwcs.yaml example.

The main thing here is that I added an option to provide an `input.wcs` field to specify a WCS other than the one in the FITS file.  To make that easier (and also, because I wasn't super happy with how the code for the input field parsing was written), I decided to leverage the galsim config parser for our purposes here and rewrite most of the input.py code.  Now you can specify any input field using the same kinds of value types and such as are allowed in galsim config files.  

In addition, you can specify a `modules` field which works the same way as it does in a galsim config file.  This let me use the Pixmappy WCS type from the `galsim_extra` module for the true WCS.

Along the way, I made a few other changes as well:

- Changed `images` to `image_file_name` in input field.  Having it be plural was confusing when there is only one file (not uncommon).  And even if there are multiple files, the singular is consistent with how this field is typically named in GalSim config stuff, so I think this will be more intuitive in general.
- Changed `cats` to `cat_file_name` for the same reasons.
- Added `maxsnr` as an input option.  If any input object has a measured S/N ratio larger than this value, then Piff adds to the weight map a constant to bring the nominal S/N down to this.  It doesn't actually add noise, so the real S/N is still high, but this prevents the interpolator from over-weighting a few very bright stars when doing the fit.
- Added `minsnr` as an input option.  If any input object has a measured S/N ratio smaller than this value, then Piff removes it from the input list.
- Added `normalize` option (defaulting to True) to GPInterp.  In particular, for sizes, the GP assumption that the mean is 0 is wildly incorrect.  I think we will thus usually want to have Piff automatically subtract the mean of the parameters before doing GP interpolation.  But the option is there to turn this off if desired.
- Allow `max_remove` for the outlier rejection to be a float between 0 and 1, in which case, it is interpreted as the maximum fraction of stars to remove.  With full focal plane fits, I was getting ~7000 stars to fit.  So my usual `max_remove=3` or `5` was way too small.  It seems more appropriate to limit to ~1% of the stars each pass rather than some fixed value in cases like this.
- When the total chisq for outlier rejection is > dof, then scale up the threshold for outlier rejection by this ratio.  This is basically indicating that the noise is underestimated by this factor, so this is the more appropriate threshold to use.  (Otherwise, you could easily end up throwing everything out.)
- Fixed how the Poisson noise is added to the weight map.  We had been adding it immediately on input, but this can cause biases because the noise level is estimated from the data.  Instead, it is better to use the model as the estimator for the Poisson noise to add.  So now we add a fresh estimate of the Poisson noise each iteration through the fitting loop.  (I actually thought we were already doing this, but when I checked I discovered that we were doing it the wrong way.)


rmjarvis:
Here is a summary of this working that I presented at the DES meeting last week:

Here is what the low-noise images look like: (constant true Moffat PSF across whole field, Pixmappy true WCS, DESDM solution in FITS file)  Also, there were 60 chips simulated, this is just 4 of them.
image

PSFEx results (rho1): (order=2)
image

Piff results using single chips, order=3, still using the FITS WCS.
image

Piff results using single chips, order=3, overriding the WCS with the true (Pixmappy) value:
image

Piff results using full focal plane, order=5, Pixmappy WCS:
image

Piff results using full focal plane, order=0, Pixmappy WCS:
image

I went on to do more realistic PSF patterns with Gaussian Processes, but I haven't really gotten that working well yet.  So that will be my next effort.

piff/input.py
```diff
@@ -176,422 +80,558 @@ def makeStars(self, logger=None):
...
+            :sky_col:       The name of a column with sky values. [default: None]
+            :gain_col:      The name of a column with gain values. [default: None]
+            :sky:           The sky level to subtract from the image values. [default: None]
+                            Note: It is an error to specify both sky and sky_col. If both are None,
+                            no sky level will be subtracted off.
+            :gain:          The gain to use for adding Poisson noise to the weight map.  [default:
+                            None] It is an error for both gain and gain_col to be specified.
+                            If both are None, then no additional noise will be added to account
+                            for the Poisson noise from the galaxy flux.
+            :minsnr:        The minimum S/N ratio to use.  If an input star is too faint, it is
```
rmjarvis: Sure.  I'm ok with adding an underscore for these.

piff/input.py
```diff
@@ -176,422 +80,558 @@ def makeStars(self, logger=None):
...
+                            removed from the input list of PSF stars.
+            :maxsnr:        The maximum S/N ratio to allow for any given star.  If an input star
```
rmjarvis: I did wrestle with this difference a bit, but I do think min and max are the right prefixes, since at the end, the stars do all fall into this range of snr.  We just do a different thing to get there at the high and low end.  We can't remove noise to lift low snr stars up to min_snr, so we have to just clip them.  But at the high end we can add noise, so no point in throwing out these stars.

Do you think there is any use case for people wanting to throw out high snr stars?  Maybe we could keep max_snr as the name, but add another option to specify whether high snr stars are clipped or have noise added.

piff/input.py
```diff
@@ -176,422 +80,558 @@ def makeStars(self, logger=None):
...
+                            is too bright, it can have too large an influence on the interpolation,
+                            so this parameter limits the effective S/N of any single star.
+                            Basically, it adds noise to bright stars to lower their S/N down to
+                            this value.  [default: 100]
+            :nstars:        Stop reading the input file at this many stars. [default: None]
```
rmjarvis: It's per input star catalog.  So typically per ccd.  I'll add some documentation to make that clearer.

piff/simplepsf.py
```diff
@@ -178,6 +178,10 @@ def fit(self, stars, wcs, pointing,
             if hasattr(self.model, 'reflux'):
                 new_stars = []
                 for s in self.stars:
+                    # Update the noise with the new interpolated signal.
+                    signal = self.drawStar(s)
+                    s = s.addPoisson(signal)
```
rmjarvis: We don't add noise to the image, just the weight map.  It takes the original variance from the input weight map and adds the (signal image / gain) to the variance.  It's better to do this with the model signal rather than the observed data (as we had been doing), since the latter can lead to biases.

tests/test_pixel.py
```diff
@@ -771,15 +780,16 @@ def test_single_image():
             },
         },
     }
-    if __name__ == '__main__':
+    if __name__ != '__main__':
+        config['verbose'] = 0
+    if True:
```
rmjarvis: I was flipping which sets of tests to include always and which to include only from the command line (not from nosetests) to try to get full test coverage while still not taking forever.  But I'm happy to get rid of the `if True` and slide things left now.

PR #58, opened 2017-08-10
Input tests

rmjarvis:
This PR adds a bunch of unit tests for the input module.  It also fixes a couple bugs they found, including (I think) one that Chris found about the dir not being parsed correctly.

It's all stuff I should have included in PR #53, but didn't...

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -317,6 +317,7 @@ class InputFiles(Input):
                 'x_col' : str,
                 'y_col' : str,
                 'sky_col' : str,
+                'gain_col' : str,
                 'flag_col' : str,
                 'use_col' : str,
                 'image_hdu' : int,
@@ -599,7 +600,7 @@ class InputFiles(Input):
             cat = cat[cat[use_col]!=0]
 
         # Limit to nstars objects
-        if nstars is not None and nstars > len(cat):
+        if nstars is not None and nstars < len(cat):
             logger.info("Limiting to %d stars for %s",nstars,cat_file_name)
             cat = cat[:nstars]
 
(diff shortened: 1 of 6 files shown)


piff/input.py
```diff
@@ -572,7 +613,7 @@ def readStarCatalog(self, cat_file_name, cat_hdu, x_col, y_col, flag_col, use_co
             cat = cat[cat[use_col]!=0]
 
         # Limit to nstars objects
-        if nstars is not None and nstars > len(cat):
+        if nstars is not None and nstars < len(cat):
             logger.info("Limiting to %d stars for %s",nstars,cat_file_name)
             cat = cat[:nstars]
```
cpadavis: (This is something I missed the previous time around, sorry.) Do we really want it to just take the first nstars? I think the sextractor catalogs at least are built up in ascending x, y coordinates so that you would just get one edge of the CCD this way. We could just do np.random.choice(len(cat), nstars, replace=False) though I dunno if we like the randomness.

tests/test_input.py
```diff
@@ -0,0 +1,673 @@
...
+    config = {
+                'dir' : dir,
+                'image_file_name' : image_files,
+                'cat_file_name': cat_files,
+                'chipnum' : [ 5, 6, 7 ]
+             }
+    input = piff.InputFiles(config, logger=logger)
+    assert input.chipnums == [ i+5 for i in range(3) ]
+
+    # Use the GalSim Eval capability to get the index + 1
```
cpadavis: Is there an ability to get the chipnum either from a column in the header or in the file name? Or is the thought that if we wanted a DES PSF's chipnums to correspond to the ccd number, we would just input the list ourselves?

rmjarvis: Sure.  I added an example of how to do it [here](62b9b98).

piff/input.py
rmjarvis: I only use the nstars feature to save time when debugging stuff.  If I'm just checking whether the code runs through to the end and does something reasonable, a lot of times it's handy to limit the number of stars being processed to only 20 or so per chip.  In which case, I don't care much which 20 it is.

I think probably if people want to select a subset of the full catalog where they would care that is not just the first N objects, they can probably add either a flag or use column and specify which ones to use that way.  We could also add a generic boolean `use` field that could be any calculation the user specifies.  But maybe I'll wait until someone actually has a concrete use case to help guide the API.

Issue #56, opened 2017-07-18
Plotify


jmeyers314:
It would be nice to be able to take an existing piff yaml + output file, modify the plots section of the yaml file, and then remake the plots but without rerunning the actual fit.  I suggest "plotify" to be a script that reads a normal looking piff yaml file, but reads the .piff file in instead of writing it out, and then also remakes the plots.

cpadavis:
one thing that you'll need to also add is the ability for stars to save their images / weight maps, and for plotify to handle both cases. I think the stats modules right now assume the stars have images / weights. Maybe we can get around that by saving the measured shapes and model shapes when we save the stars, since I think all of the stats modules right now work entirely in that space

rmjarvis:
The idea is that it would read in the input files again according to the input field in the config file.  But then rather than running the solver to make the psf, it would read in file that is listed in the output field.  I don't think we want to be writing out all the images into the .piff file just for this kind of thing.

rmjarvis:
Done.  PR #59.  Thanks Chris!

PR #59, opened 2017-08-11
Plotify

cpadavis:
Addresses issue #56 . Very simply, you can execute "plotify config.yaml" to load up a set of images, PSF file, and run stats modules on them. This means you could, for example, load a _different_ image or catalog (e.g. a validation catalog) and see the PSF's extrapolation.

It figures out the file name from config['output']['file_name'] and config['output']['dir'] (if given). Test is done in test_simple.py

(Not sure why I had some residual improvements to 2d stats hanging around my directory. Those are innocuous.)


cpadavis:
pingping -- is there any reason not to merge this?

tests/test_simple.py
```diff
@@ -364,6 +364,51 @@ def test_single_image():
     p = subprocess.Popen( [piffify_exe, 'simple.yaml'] )
     p.communicate()
 
+    # repeat using the plotify executables
```
rmjarvis: This test function is really getting rather unwieldy.  Can we please break it at the point where you start testing all the plots.  i.e. the comment `# Test that we can make rho statistics`.  At that point, it really changes focus to testing plots. So from there onward (including this new material) should be a new function called `test_plots` or some such.  You can pull out whatever code you need above that into a helper function if necessary to avoid too much code duplication.

tests/test_twodstats.py
```diff
@@ -136,3 +190,4 @@ def generate_starlist(n_samples=500):
 if __name__ == '__main__':
     test_twodstats()
     # yaml test is in test_simple
```
rmjarvis: Indeed, perhaps this comment is tell us that once you break up `test_single_image` into two, the second half should be moved to this file.

tests/test_simple.py
```diff
@@ -364,6 +364,51 @@ def test_single_image():
...
+    config['verbose'] = 0
+    with open('simple.yaml','w') as f:
+        f.write(yaml.dump(config, default_flow_style=False))
+    p = subprocess.Popen( [plotify_exe, 'simple.yaml'] )
+    p.communicate()
+
+    # test running plotify with dir in config and with no logger
+    config['output']['dir'] = '.'
+    os.remove(rho_psf_file)
+    os.remove(shape_psf_file)
```
rmjarvis: I think these should be identical each time.  Maybe we could use `diff` to check that the pdf files are identical rather than just delete them each time.  Then there would be at least a bit of a test besides just checking that no errors occurred.

cpadavis:
It looks like you did my homework for me -- sorry, and thanks!

rmjarvis:
No worries.  :)  
There was some other cleanup I wanted to do on some files as well, so this branch seemed like a good place to do it.  Mostly I wanted to speed up the unit tests, although I seem to have lowered the test coverage somewhat, so now I'm looking into that.

Issue #55, opened 2017-07-18
Model image plots


jmeyers314:
It would be nice to see the images for our model PSFs along side star (data) images and residuals.

rmjarvis:
Done.  PR #60.

PR #60, opened 2017-08-12
Plot stars

cpadavis:
(I dunno if breaking these chunks into little bits is too much, but since these are easy fixes I thought I would do it this way.)

This addresses issue #55 . This PR adds a StarStats object which plots stars, PSF, and residual. You can specify the number of stars. If that number is less than the total number of stars, then it will randomly select without replacement that number of stars for plotting. (Don't worry -- it tells you which stars it uses in the plot so you can recover it.) By default it will create a figure with 5 such stars plotted.

The tests for it live in test_simple.py.

### DIFF
--- a/piff/__init__.py
+++ b/piff/__init__.py
@@ -103,6 +103,7 @@ from .singlechip import SingleChipPSF
 # Stats classes are named BlahStats where Blah is what they are called in the config file
 from .stats import Stats, RhoStats, ShapeHistogramsStats
 from .twod_stats import TwoDHistStats, WhiskerStats
+from .star_stats import StarStats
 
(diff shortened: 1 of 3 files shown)


cpadavis:
pingping -- is there any reason not to merge this?

rmjarvis:
Sorry.  I forgot about these.  I'm on vacation in NZ this week, but I'll try to get to reviewing them soon after I get back.

rmjarvis:
Looks very useful Chris.  I'm looking forward to trying this one out on some of the problematic fields in the Y3 data to try to help diagnose how things are going wrong.

PR #64, opened 2017-11-16
Gp 2pcf

PFLeget:
Hi, 

I put within piff the used of two-point correlation function to fit hyperparameters and compute the interpolation. I presented 3 weeks ago at the last piff meeting. To do that, I created a specified script, gp_interp_2pcf.py that is similar to the script that used the scikit-learn gaussian process, but I am using scikit-learn just for kernel purpose. I am doing the final interpolation by myself, without using scikit-learn or george. 

So this version of  gp_interp_2pcf.py is different from gp_interp.py on different points:

- Used two-point correlation function instead of maximum likelihood to find hyperparameters
- For each parameter, a different search of hyperparameter is done.
- You can choose different kernel for all the different parameter (allow also to start with different initial guess)
- The error from lmfit is propagated to the interpolation 
- The final interpolation is compute directly within the code using cholesky decomposition from scipy instead of doing that within scikit-learn

Moreover, I created a new script called kernel.py where I put all the previous kernels wrote within Piff (like the AnisotropicRBF kernel) and I added the VonKarman one that I used in the simulation that I shown 3 weeks ago. Also, I added new test for all those new implementation (test_gp_interp_2pcf.py)

If you have any question do not hesitates.
 
Cheers, 
Pierre-Francois


piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+
+from sklearn.gaussian_process.kernels import Kernel
+import scipy.optimize as op
+from scipy.linalg import cholesky, cho_solve
+
+from .interp import Interp
+from .star import Star, StarFit
+
+
+class GPInterp2pcf(Interp):
```
cpadavis: So based on the fact that this is a sort of specialization of GPInterp, I am wondering if it shouldn't inherit from that instead. Maybe you have to change enough functions that this is clearer, but as I go through here I do find myself wondering whether some of the changes here are backported back to GPInterp.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+        :param kernel: sklearn.gaussian_process kernel.
+        :param y_err: Error of y. (n_samples, n_targets)
+        """
+
+        if y_err is None:
+            y_err = np.zeros_like(y)
+        HT = kernel.__call__(X2,Y=X1)
+        K = kernel.__call__(X1) + np.eye(len(y))*y_err**2
+        factor = (cholesky(K, overwrite_a=True, lower=False), False)
+        alpha = cho_solve(factor, y, overwrite_b=False)
```
cpadavis: Based on the way this is written, I think the majority of times this function is called it will be using the same X1, y, y_err from when the model is fitted. This means you wouldn't have to keep recomputing alpha, which probably would help speed things up.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+                fit = star.fit.newParams(y0)
+            fitted_stars.append(Star(star.data, fit))
+        return fitted_stars
+
+    def _finish_write(self, fits, extname):
+        # Note, we're only storing the training data and hyperparameters here, which means the
+        # Cholesky decomposition will have to be re-computed when this object is read back from
+        # disk.
+
+        init_theta = np.array([self._init_theta[i] for i in range(self.nparams)])
```
cpadavis: is there a reason for keeping the initial theta?

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def solve(self, stars=None, logger=None):
+        """Set up this GPInterp object.
+
+        :param stars:    A list of Star instances to interpolate between
+        :param logger:   A logger object for logging debug info. [default: None]
+        """
+        X = np.array([self.getProperties(star) for star in stars])
+        y = np.array([star.fit.params for star in stars])
+        y_err = np.array([np.sqrt(np.diag(star.fit.params_cov)) for star in stars])
```
cpadavis: not all models produce params_cov. This should probably check that stars have params_cov, and otherwise fill with zeros

piff/kernel.py
```diff
@@ -0,0 +1,301 @@
...
+        if eval_gradient:
+            raise RuntimeError("Cannot evaluate gradient for ExplicitKernel.")
+
+        X = np.atleast_2d(X)
+        if Y is None:
+            Y = X
+
+        # Only writen for 2D covariance at the moment
+        xshift = np.subtract.outer(X[:,0], Y[:,0])
+        yshift = np.subtract.outer(X[:,1], Y[:,1])
```
cpadavis: this should be an easy fix:

```python
shift = [np.subtract.outer(X[:,i], Y[:,i]) for i in X.shape[1]]
return self._fn(*shift)
```

piff/kernel.py
```diff
@@ -0,0 +1,301 @@
...
+        self._fn = self._eval_fn(fn)
+        self.theta = []
+
+    @staticmethod
+    def _eval_fn(fn):
+        # Some potentially useful imports for the eval string.
+        import math
+        import numpy
+        import numpy as np
+        return eval("lambda du,dv:" + fn)
```
cpadavis: this could be upgraded to arbitrary numbers of dimensions / arbitrary input variable names if you required the lambda bits to be a part of the fn string as well. This would also let you put in predefined functions which may not have du, dv as a part of them

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+
+from .interp import Interp
+from .star import Star, StarFit
+
+
+class GPInterp2pcf(Interp):
+    """
+    An interpolator that uses two-point correlation function and gaussian process to interpolate a single surface.
+
+    :param keys:        A list of star attributes to interpolate from
```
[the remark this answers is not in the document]
cpadavis: well I guess it could be done for arbitrary numbers of dims, but treecorr needs it to be 1, 2, or 3, and we would need to be able to handle len(keys) != 2

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+        :param y_err: Error of y. (n_samples, n_targets)
+        """
+
+        if y_err is None:
+            y_err = np.zeros_like(y)
+        HT = kernel.__call__(X2,Y=X1)
+        K = kernel.__call__(X1) + np.eye(len(y))*y_err**2
+        factor = (cholesky(K, overwrite_a=True, lower=False), False)
+        alpha = cho_solve(factor, y, overwrite_b=False)
+        return np.dot(HT,alpha.reshape((len(alpha),1))).T[0]
```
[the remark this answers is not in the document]
cpadavis: oh I see that this was a conscious choice on your part. I'm not really sure why this is better?

piff/gp_interp_2pcf.py
PFLeget: I guess, I changed to much things to make the possibility that GPInterp2pcf inherit from GPInterp. If I would changed only the strategy of searching hyperparameter, It would be ok. However, the fact that I am fitting different hyperparameter(s) per PSF’s parameter(s) changes a lot of thing within the class. That should be backported to GPinterp.

piff/gp_interp_2pcf.py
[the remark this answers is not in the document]
PFLeget: Indeed this totally right. I added a ValueError if the len(key)!=2. I added that in the documentation also.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+
+class GPInterp2pcf(Interp):
+    """
+    An interpolator that uses two-point correlation function and gaussian process to interpolate a single surface.
+
+    :param keys:        A list of star attributes to interpolate from
+    :param kernel:      A string that can be `eval`ed to make a
+                        sklearn.gaussian_process.kernels.Kernel object.  The reprs of
+                        sklearn.gaussian_process.kernels will work, as well as the repr of a
+                        custom piff VonKarman object.  [default: 'RBF()']
```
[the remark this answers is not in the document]
PFLeget: I did few test and if you write just ’RBF’ it will execute, but without initializing the kernel, which is a source of error if you are doing something like that right now. However if you write ‘2 *RBF’ it will not worked and you will get an error message. I don’t think it should work like that. I mean I guess this is better to feed the gp with initial guess of hyperparameter like it is done during all the test and so write you kernel with something like ‘RBF(2)’. I added a TypeError within the init in order to not allowed something like ‘RBF’ or ‘VonKarman’ alone.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+    :param optimize:    Boolean indicating whether or not to try and optimize the kernel by
+                        computing the two-point correlation function.  [default: True]
```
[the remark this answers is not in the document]
PFLeget: Like you said in the next comment, you set your initial hyperparameter(s) when you write the kernel. I guess it should be more obvious in the documentation. Instead of putting per default ‘RBF()’, I put ‘RBF(1)’. It is exactly the same thing, but I guess in the point of view of the user, this is more obvious that you can modify the value by yourself.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+            logger.debug('After fit: kernel = %s',kernel.set_params())
+
+        return kernel
+
+    
+    def _optimizer_2pcf(self,kernel,X,y,y_err):
+        """Fit hyperparameter using two-point correlation function.
+
+        :param kernel: sklearn.gaussian_process kernel.
+        :param X:  The independent covariates.  (n_samples, n_features)
```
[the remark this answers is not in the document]
PFLeget: I change it also in other functions.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+            var = disp**2
+            return np.sum(residual**2/var)
+
+        p0 = kernel.theta
+        results_fmin = op.fmin(chi2,p0,disp=False)
+        results_bfgs = op.minimize(chi2,p0,method="L-BFGS-B")
+        results = [results_fmin, results_bfgs['x']]
+        chi2_min = [chi2(results[0]), chi2(results[1])]
+        ind_min = chi2_min.index(min(chi2_min))
+        results = results[ind_min]
```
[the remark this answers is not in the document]
PFLeget: When I was doing tests, I saw that there is moment where one method is better that the other and reciprocally. This is so fast to do both minimization that I kept both. I can put the L-BFG-S as a parameters, but I guess to keep the other method is just a safety check which cost nothing for the moment.

piff/gp_interp_2pcf.py
PFLeget: I guess this is true only if you are using the same kernel and if you have exactly same hyperparameters (so which means that you are not optimizing your hyperparameters). In this case this is the same alpha for all PSF parameters. However, in the general case you have one alpha per PSF parameter.

piff/gp_interp_2pcf.py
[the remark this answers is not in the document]
PFLeget: This is within the gaussian process regressor of sklearn that they put this functionality and yes it was totally a conscious choice from my side to do something like that. And if you look the sklearn/george gp they are doing exactly that. But I choose to do something like that because I was thinking that maybe in the future, we would like to put a better/faster matrix inversion (like TMV). By not using sklearn or george for that part you are free to do what you want in the future even putting them back.

piff/gp_interp_2pcf.py
PFLeget: I add per default in StarFit within star.py a params_cov wich is set to None per default. Within the test it was working like that, but I guess it should be an output of each fitting method, not only lmfit.

piff/gp_interp_2pcf.py
PFLeget: It was to be consistent with the gp_interp.py, I have not a strong opinion on that. I guess like that you can write it explicitly in output file and keep a track of what was the input.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+        else:
+            if len(self.kernel_template)!= self.nparams:
+                raise ValueError("numbers of kernel provided should be 1 (same for all parameters) or " \
+                "equal to the number of params (%i), number kernel provided: %i"%((self.nparams,len(self.kernel_template))))
+            else:
+                self.kernels = [copy.deepcopy(ker) for ker in self.kernel_template]
+
+        for i in range(self.nparams):
+            self.kernels[i] = self.kernels[i].clone_with_theta(fit_theta[i])
+            self.optimizer = self._optimizer_init
```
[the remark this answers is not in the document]
PFLeget: It is not, I removed it.

piff/kernel.py
PFLeget: This part was originally wrote by Josh I guess. I just moved it from gp_interp.py to kernel.py. I fond that more convenient to write a separate script where everybody can implement a new kernel in a sklearn style.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+                raise TypeError("kernel should be a string or a list of string")
+            else:
+                self.kernel_template = [self._eval_kernel(ker) for ker in kernel]
+
+        self._2pcf = []
+        self._2pcf_dist = []
+        self._2pcf_fit = []
+
+        self.count = 0
+        self._optimizer_init = copy.deepcopy(optimize)
```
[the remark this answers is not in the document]
PFLeget: It is something that I implemented at the beginning for my own test and I forgot to remove it. I removed it.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,375 @@
...
+                out.append(c)
+                out.extend(recurse_subclasses(c))
+            return out
+        clses = recurse_subclasses(Kernel)
+        for cls in clses:
+            module = __import__(cls.__module__, globals(), locals(), cls)
+            execstr = "{0} = module.{0}".format(cls.__name__)
+            exec(execstr, globals(), locals())
+
+        from numpy import array
```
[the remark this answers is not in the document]
PFLeget: I have no idea … I removed it.

PFLeget:
Hi, 

I took into account comments from Chris and I fixed all Travis bug and code coverage on those last commits. If you have any comments on those last updates or on the previous commits, they are welcome.

Cheers,
Pierre-François

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,372 @@
...
+        clses = recurse_subclasses(Kernel)
+        for cls in clses:
+            module = __import__(cls.__module__, globals(), locals(), cls)
+            execstr = "{0} = module.{0}".format(cls.__name__)
+            exec(execstr, globals(), locals())
+
+        try:
+            k = eval(kernel)
+        except:
+            raise RuntimeError("Failed to evaluate kernel string {0!r}".format(kernel))
```
rmjarvis: A couple things about this.  
First, I've learned that it's unhelpful to just swallow exceptions that get raised.  At the very least, you should let the user know what the original exception was.
Second, if the exception is a KeyboardInterrupt, this isn't what you want.  So it is good practice whenever you catch all exceptions to first catch KeyboardInterrupt and reraise.  Apparently SystemExit is another one you should raise, but I don't know that I've ever seen a use of that one.  (cf. http://effbot.org/zone/stupid-exceptions-keyboardinterrupt.htm)

So this should become:
```
try:
    k = eval(kernel)
except (KeyboardInterrupt, SystemExit):
    raise
except Exception as e:
    raise RuntimeError("Failed to evaluate kernel string {0!r}.  Original exception: {1}".format(kernel, e))
```
Oh, and this isn't the only place where this pattern exists in Piff, so if you don't mind copying over this pattern to the other places where we were sloppy, that would be much appreciated.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,372 @@
...
+        :param y:  The dependent responses.  (n_samples, n_targets)
+        :param y_err: Error of y. (n_samples, n_targets) [default: None]
+        :param logger:  A logger object for logging debug info. [default: None]
+        """
+        if logger:
+            logger.debug('Start kernel: %s', kernel.set_params())
+            logger.debug('gp.fit with mean y = %s',np.mean(y))
+        # Save these for potential read/write.                
+        if y_err is None:
+            y_err = np.zeros_like(y)
```
rmjarvis: This branch is never tested.  Nor the other functions where you allow y_err to be None.  I think that given how `solve` is written, it's not possible to get y_err=None.  So probably remove that default and option.  Else add a test that uses `y_err=None`.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,372 @@
...
+            y = self._pca.transform(y)
+            if y_err is not None:
+                y_err = self._pca.transform(y_err)
+            self._y_pca, self._y_pca_err = y, y_err
+            self.nparams = self.npca
+
+        if self.normalize:
+            self._mean = np.mean(y,axis=0)
+        else:
+            self._mean = np.zeros(self.nparams)
```
rmjarvis: Should add a test with `normalize=False`.  This branch is currently untested.

piff/kernel.py
```diff
@@ -0,0 +1,293 @@
...
+        self.length_scale = length_scale
+        self.length_scale_bounds = length_scale_bounds
+
+    @property
+    def anisotropic(self):
+        return np.iterable(self.length_scale) and len(self.length_scale) > 1
+
+    @property
+    def hyperparameter_length_scale(self):
+        if self.anisotropic:
```
rmjarvis: You don't seem to test this option.  Should add a test that hits these branches.

piff/kernel.py
```diff
@@ -0,0 +1,293 @@
...
+        if Y is None:
+            dists = pdist(X, metric='euclidean')
+            K = (dists**(5./6.)) * ssp.kv(-5./6.,2*np.pi*dists/length_scale)
+            K = squareform(K)
+
+            dd_w = np.linspace(1e-4,1,100) / length_scale
+            dd = np.linspace(1e-4,1,100)
+            spline = inter.InterpolatedUnivariateSpline(dd,
+                                                        (dd**(5./6.)) * ssp.kv(-5./6.,2*np.pi*dd_w))
+            np.fill_diagonal(K, spline(0))
```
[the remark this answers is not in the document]
PFLeget: I don’t think this is correct in this context. The equation that I implemented is  r^5/6 * BesselK(5/6, 2 * pi * r / L) where L is the correlation length. By changing the value of L, the value of the function at r=0 changes. However, I guess if you have an analytical solution for r^5/6 * BesselK(5/6,r), I guess this is easy for r^5/6 * BesselK(5/6, 2 * pi * r / L) but I am not able to find it. How did you do it for r^5/6 * BesselK(5/6,r)?

Thanks in advance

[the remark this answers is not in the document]
rmjarvis: I had Maple do it for me, but [this page](http://functions.wolfram.com/Bessel-TypeFunctions/BesselK/introductions/Bessels/05/) for instance has the Taylor expansion of K_nu(z), from which you can do it yourself.  I guess in this case, you need to multiply the dimensionless result by (L/2pi)^5/6.  
You can probably verify the result by adding print statements to compare what your code gets for `spline(0)` with `0.5 * (length_scale/np.pi)**(5./6.) * spacial.gamma(5./6.)` and make sure they match to some level of expected numerical precision.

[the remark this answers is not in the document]
jmeyers314: wolframalpha is also useful for these sorts of calculations: 

http://www.wolframalpha.com/input/?i=limit%5Br%5E(5%2F6)+BesselK%5B5%2F6,+2+pi+r+%2F+L%5D,+r-%3E0%5D

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,372 @@
...
+
+"""
+.. module:: gp_interp_2pcf
+"""
+
+import numpy as np
+import treecorr
+import copy
+
+from sklearn.gaussian_process.kernels import Kernel
```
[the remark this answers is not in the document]
PFLeget: Indeed. Even if george got a better API than sklearn gaussian process, this is easiest to implement new kernel with sklearn like Josh did for the Anisotropic RBF kernel and like I did for the VonKarman kernel. Now the search of hyperparameter(s) is done explicitly outside sklearn/george, so there is no a main interest to used one or the other in this configuration. And by keeping only sklearn kernel, it stays consistent with the other gp_interp.py that used maximum likelihood.

piff/gp_interp_2pcf.py
PFLeget: I implemented this within gp_interp.py and gp_interp_2pcf.py.

piff/gp_interp_2pcf.py
PFLeget: Yes, I agree. I put in StarFit the variance instead of the covariance, and in gsobject, I return the variance instead of the full covariance matrix. For PixelGrid, and optical_model, I set the variance for the moment to zeros.

piff/gp_interp_2pcf.py
```diff
@@ -0,0 +1,372 @@
...
+        y = np.array([star.fit.params for star in stars])
+        y_err = np.array([np.sqrt(np.diag(star.fit.params_cov)) for star in stars])
+
+        self._X = X
+        self._y = y
+        self._y_err = np.sqrt(y_err**2 + self.white_noise**2)
+
+        if self.npca > 0:
+            from sklearn.decomposition import PCA
+            self._pca = PCA(n_components=self.npca, whiten=True)
```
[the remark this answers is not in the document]
PFLeget: Yes I agree. I did it like that in order to have something consistent with gp_interp.py. However, I already did something like that using Factor Analysis (like PCA, except if you have noise). Doing something like that provide you Normalized vector, and all the variance would be in explonary factor instead of the vector size as far as I understand the documentation. I guess this is more an arbitrary choice. I removed it for the moment in both gaussian process implementation, but if somebody thinks it is useful, we can add it again.

piff/kernel.py
PFLeget: I do not used it for the moment, because I want to use it latter, when I will add option of fitting anisotropic kernel with two-point correlation function. Shall I remove it or can I let it like that for the moment ?

piff/kernel.py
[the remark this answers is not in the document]
PFLeget: That’s great, the solution provided by this website is perfect :D

tests/test_gp_interp.py
```diff
@@ -211,6 +211,58 @@ def make_grf_psf_params(ntrain, nvalidate, nvisualize):
...
+    ntotal = ntrain + nvalidate + nvisualize**2
+    params = np.recarray((ntotal,), dtype=star_type)
+
+    np.random.seed(30352010)
+    
+    # Training
+    #us = [ud() for i in range(ntrain)]
+    #vs = [ud() for i in range(ntrain)]
+    us = [np.random.uniform(0,1) for i in range(ntrain)]
+    vs = [np.random.uniform(0,1) for i in range(ntrain)]
```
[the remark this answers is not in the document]
PFLeget: There is no reason, I wanted to look something with numpy and I forgot to remove it. I set like before. Yes indeed, I just try to reproduce what Josh did for RBF kernel.

PFLeget:
Hi,

I guess that I implanted all the comments from Mike. If you have any comments on those last updates or on the previous commits, they are welcome.

Cheers,
Pierre-François

piff/kernel.py
```diff
@@ -0,0 +1,290 @@
...
+
+        X = np.atleast_2d(X)
+        length_scale = _check_length_scale(X, self.length_scale)
+        if Y is None:
+            dists = pdist(X, metric='euclidean')
+            K = (dists**(5./6.)) * special.kv(5./6.,2*np.pi*dists/length_scale)
+            K = squareform(K)
+
+            dd_w = np.linspace(1e-4,1,100) / length_scale
+            dd = np.linspace(1e-4,1,100)
```
rmjarvis: I think you don't need dd, dd_w anymore.

PFLeget: Yes Indeed, this is useless now. I fixed it. 

However, which is totally weird is the fact that now it does not pass Travis Cl tests for python 2.7 and 3.5 (works for python 3.4). And on my own laptop tests are working. By looking error message, it seems it is not connected with my new implementation, but with the fact that it is not able to build the environnement to start the test (fail after 2 minutes of running on Travis Cl). I tried to restart the Travis test for python 2.7 and python 3.5 few times. By the past it already happened once, and just to restart the tests worked. But it does not work now… I am not to much familiar with Travis Cl, do you have an idea of how to fixed that?

Thanks a lot

rmjarvis: Usually that means something changed about the Travis environment.  I'll see if I can figure it out.

rmjarvis:
I just added something that would help diagnose what the error was, and it fixed itself, so I think it was just something wonky in the cache that needed to be reset to work.  Anyway, I think we're good to go here.

PR #65, opened 2018-02-15
Von-Karman kernel equal to 1 when distance equal to 0

PFLeget:
Hi everyone, 

I made a little change on the Von-Karman kernel. Before this modification the value of this kernel for a distance equal to zeros was equal to:

special.gamma(5./6.) /(2 * ((np.pi / length_scale)**(5./6.)) )

The problem with that is just if you change the correlation length, it does change the value of the correlation function for a distance equal to zeros. This is a problem when you are multiplying the Von-Karman kernel by a constant Kernel, because basically you will not have the same constant with different correlation length, even if the true standard deviation of your data (like PSF size) does not change. 

To fixed that, I modified a bit the definition of the Von-Karman kernel in order to get a value of the kernel equal to one for a distance equal to zeros and for any correlation length. 

I guess this is better like that if we want to get a physical interpretation of the constant kernel (the variance of the data) when it is multiply by the Von-Karman kernel. 

If you have any question do not hesitates.
Cheers,
Pierre-François

### DIFF
--- a/piff/kernel.py
+++ b/piff/kernel.py
@@ -258,7 +258,7 @@ class VonKarman(StationaryKernelMixin, NormalizedKernelMixin, Kernel):
 
             dists = cdist(X, Y, metric='euclidean')
             K = ((dists/length_scale)**(5./6.)) * special.kv(5./6.,2*np.pi*dists/length_scale)
-            Filter = np.isfinite(K)
+            Filter = (dists != 0.)
             lim0 = special.gamma(5./6.) / (2 * (np.pi**(5./6.)))
             if np.sum(Filter) != len(K[0])*len(K[:,0]):
                 K[~Filter] = lim0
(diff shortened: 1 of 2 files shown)


jmeyers314:
Hi PF, all the changes look good to me, and I agree that this is a better parameterization for a VonKarman kernel.

I did have one question about the pre-existing code though.  [Here](https://github.com/rmjarvis/Piff/blob/5503c1998dd02d990765d35fe35dbabb5ce3ba1f/piff/kernel.py#L261) , would it be clearer to write `Filter = (dists==0)`?  If there are other reasons why `K` could be infinite, we probably don't want to just set those values to `lim0`.  Is that right?

PFLeget:
@jmeyers314 , Yes indeed this is right. I added this recommendation within the Von-Karman kernel, because I guess it make it more clear. 

Let me know if I can merge this pull request. 

Cheers,
PF

Issue #66, opened 2018-02-28
Average of PSF's parameters !=0 across different exposures & GP interp


PFLeget:
Hi, 

I am working right now on atmospheric fit from DES Y1 images. By looking PSF’s parameters averages across 23 exposures, I notice (like @cpadavis ), that you can observed a systematic average, different from zeros and that vary across the FoV. Some of those average fluctuation are correlated to some CCD position. Other looks like rings pattern, which I guess is something known from SV and Y1 data and solved for Y3 with better astrometric solution. 

Even if those problem disappear in Y3, I expect to have this same kind of problem (average across exposures !=0), and this is an issue for both Gaussian Process algorithms implemented within Piff. Indeed, the assumption of the GP is that the gaussian fluctuation are around an average function that can be a function of the coordinates (which is what I observed on Y1 data). So, not take into account this average will affect, the shape of the spatial correlation, the hyperparameters found and the quality of the interpolation.

This is not a big deal to treat that correctly, it just needs to subtract this average function determine across different exposure on a single exposure before to start the interpolation and to add it back when you get your GP interpolation. 

I would like to implement this new features within both GP algorithm that are implemented within Piff. Basically to do that, I want to feed the GP class with a new argument that are a function of the coordinate and which return the average of the PSF’s parameters. However, I know that Piff should work also from a config files so my idea is to do something like that:

Using the dill package (which is pip installable and equivalent to pickle) you can wrap a function/class within a pickle files and after re-used the function by opening the pickle file. An example: 

```python
import dill 

class average:
	def __init__(self, coeff_x, coeff_y):
		self.coeff_x = coeff_x
		self.coeff_y = coeff_y

	def __call__(self, x, y):
		return self.coeff_x * x + self.coeff_y * y

mean = average(2, 3)
print mean(5, 2)
dill.dump(mean, open("test.pickle", mode="wb"))
```

and now from everywhere this is possible to re-used this function with coefficient 2 and 3:

```python
import dill 
mean  = dill.load(open("test.pickle", mode="rb"))
print mean(5, 2)
```

I guess with this implementation, you just need to feed the config file with this pickle file and within the GP class, used directly the function implemented within the pickle.
Do you think this is a good API to do it like that?

I am writing a new branch to implement this spatial average define in a pickle file. If you think, it should be done in an other way, comments are welcome. But I am convince at least, GP implementation should care about average across exposures because this is something needed to make GP interpolation work on real data.

Cheers,
PF

[...]

PR #70, opened 2018-04-04
Subtract spatial average gp

PFLeget:
Hi,

As discussed in Issue #66 , I added within piff the possibility of computing the spatial average function of PSF parameters, and to use it for the gaussian process interpolation in gp_interp_2pcf.py. 

I wrote in config.py a meanify function (with an associated script) that takes piff outputs, computes the average of the different psf parameters and writes the average function in a fits file. Once it is done, it is possible to feed gp_interp_2pcf.py with it. 

I wrote tests of those new functionality in test_meanify.py

I guess this possibility of computing average function of PSF’s parameters could be useful, in order to deal with systematic residuals of the optical model, when atmospheric parameter are interpolated. 

If you have any questions do not hesitate.

Cheers,
Pierre-Francois


cpadavis:
it looks like travis-ci ran `test_gp_meanify` before` test_meanify` and therefore failed. The former needs the output of the latter. A cursory look at nosetests docs says:

> nose runs functional tests in the order in which they appear in the module file. TestCase-derived tests and other test classes are run in alphabetical order.

I'm not sure what the distinction is between functional tests and TestCase-derived tests. Since `test_meanify` appears before `test_gp_meanify`, maybe nosetests runs TestCase type tests? The hacky way might be to rename the tests to something like `test_1_meanify` and `test_2_gp_meanify`, or some other renaming like that.

rmjarvis:
> The hacky way might be to rename the tests to something like test_1_meanify and test_2_gp_meanify, or some other renaming like that.

Please don't be hacky.  The right way to do it is to write a `setup()` function for nosetests to run before any of the `test*` functions.  You can look at test_stats.py for an example of this pattern.

piff/config.py
```diff
@@ -191,3 +191,122 @@ def plotify(config, logger=None):
     for stats in output.stats_list:
         stats.compute(psf,stars,logger=logger)
         stats.write(logger=logger)
+
+def meanify(config, logger=None):
+    """Take Piff output(s), build an average of the FoV, and wirte output average.
```
rmjarvis: write

piff/config.py
```diff
@@ -191,3 +191,122 @@ def plotify(config, logger=None):
...
+            raise ValueError("%s field is required in config dict output"%key)
+
+    if 'dir' in config['input']:
+        dir = config['input']['dir']
+    else:
+        dir = None
+
+    if 'binning' in config['output']:
+        if type(config['output']['binning']) is not int:
+            ValueError("type of binning should be integer (%s now)"%type(config['output']['binning']))
```
rmjarvis: "type of binning" being an integer seems incongruous.  Probably this should be `nbins`?  

Or, probably better, this should be given as bin_spacing in arcsec, rather than nbins.  People may run this over single ccds or full fields or even partial fields probably.  Also, not all possible fields are the same size in x and y (e.g. if running on single chips -- they could be 2K x 4K.)  So specifying a physical size of the cells in which to average things is probably more useful than specifying a number of bins across the field.

piff/config.py
```diff
@@ -191,3 +191,122 @@ def plotify(config, logger=None):
...
+            raise ValueError("file_name may not be an empty list")
+    elif isinstance(config['input']['file_name'], str):
+        file_name = config['input']['file_name']
+        if dir is not None:
+            file_name = os.path.join(dir, file_name)
+        psf_list = sorted(glob.glob(file_name))
+        if len(psf_list) == 0:
+            raise ValueError("No files found corresponding to "+config['file_name'])
+    elif not isinstance(config['file_name'], dict):
+        raise ValueError("file_name should be either a dict or a string")
```
rmjarvis: I don't particularly like the file name structure here.  

The way you have it set up, the user will first run a regular config file first with `input` being the data and `output` being the .piff files.

Then they need to rearrange things so that `input` is the old `output` and `output` is a new file for the averages.

Then the do a third run that looks mostly like the first run, but now has an additional `model.average_fits` field.

I think I'd rather let the user use the same config file or at least as close as possible, in all three cases.  And even better if we can combine the first two steps into one.

I'm thinking maybe have a new top-level field, not in output or model or anything, called `hyper`, which would store hyper parameters that use information from multiple exposures.  Then `meanify` would look for an existing set of output files, make them if not present (using the normal piffify function), and then write out to `hyper.file_name` as an output.
Then the GP stuff would look for a `hyper` section and if there is one, get the mean parameters from there to subtract off.

The one thing I think we'd need to add to the regular processing is the ability to specify multiple output files to run (and correspondingly multiple input exposures).  That's not too hard using the GalSim config processing stuff that we are already using to loop over multiple images.  But you don't need to try to do that.  I'll work on that on a separate branch.

For now, I think, just rearrange what you have so that the input to meanify is the `output` field, since that's where it will be in the normal config file.  And its output file name is in a `hyper` field.  Then have that be where GPInterp2pcf will look for the name of the average file.

Then I can tackle making this a bit more seamless for the user to build and run on a set of exposures.

piff/gp_interp_2pcf.py
```diff
@@ -253,6 +275,20 @@ def initialize(self, stars, logger=None):
                 self.kernels = [copy.deepcopy(ker) for ker in self.kernel_template]                                        
         return stars
 
+    def _build_average(self, X):
```
cpadavis: for readability it might help to name this something that mentions it is an average from _prior_ information. I initially read this and was confused because it reads as being an average from the input data.

piff/config.py
PFLeget: I totally agree for that. However, the coordinate u & v, are they all the time in arcsec ? If yes, there is no problem for me to implement the binning in physical units.

rmjarvis: Yes, u,v are defined to be in arcsec.  (Unless someone inputs a pathological WCS I suppose, but let's not worry about that.)

piff/config.py
PFLeget: Ok, for the moment like Mike suggests, I replaced the `input` field by `output`, and the `output` field by `hyper`

PFLeget:
Hi everyone, 

I did implement all the changes requested. Let me know if you have new comments with those new changes. 

Cheers, 
PF

piff/config.py
```diff
@@ -210,36 +207,33 @@ def meanify(config, logger=None):
...
-    if 'binning' in config['output']:
-        if type(config['output']['binning']) is not int:
-            ValueError("type of binning should be integer (%s now)"%type(config['output']['binning']))
-        else:
-            binning = config['output']['binning']
+    if 'bin_spacing' in config['hyper']:
+        bin_spacing = config['hyper']['bin_spacing'] #in arcsec
     else:
-        binning = 15
+        bin_spacing = 120. #default bin_spacing: 120 arcsec
```
rmjarvis: 20 arcmin seems like a lot.  I'd have guessed more like 5 arcmin maybe? A single CCD in DES is 8x18 arcmin, for reference.  LSST is similar: 14 x 14 arcmin.

PFLeget: I guess that I put 120 arcsec per default which are 2 arcmin, but maybe there is something that I am misunderstanding.

rmjarvis: Oh, that was dumb.  You're right of course.  Somehow I kept thinking that was a factor of 10 larger than it was.  120 as the default is fine.

PFLeget:
Hi everybody, 

If nobody have new suggestions/comments, I will merge this branch next Thursday.

rmjarvis:
I rebased it to master, so it should be clean to merge now.  But if you want to add any more commits, you'll want to `git reset --hard origin/subtract_spatial_average_gp` before you make any local changes.

PR #77, opened 2018-09-05
Improve meanify

PFLeget:
Hi everyone, 

I added new options into meanify. Now you have the option of using the median. Basically, I added two options in the config of meanify within the `'hyper'` key. You can choose your desired statistics now (`'mean'` or `'median'` and `'mean'` is default). The other option is a list of integer `'params_fitted'` which indicates on which parameters you want to compute the `'mean'`/`'median'`. I added this last option because computing the median is really slow in comparaison of the mean, and I don’t need to compute the mean for all PSF’s parameters in the case of the OptAtmo model (only for atmosphere). 

It seems that the median improves a little bit the Rowe statistics on what I saw. I can show you some results on that at the next Piff meeting. 

If you have any questions do not hesitate.
Cheers,
Pierre-Francois

### DIFF
--- a/piff/config.py
+++ b/piff/config.py
@@ -229,6 +229,15 @@ def meanify(config, logger=None):
     else:
         bin_spacing = 120. #default bin_spacing: 120 arcsec
 
+    if 'statistic' in config['hyper']:
+        if config['hyper']['statistic'] not in ['mean', 'median']:
+            raise ValueError("%s is not a suported statistic (only mean and median are currently suported)"
+                             %config['hyper']['statistic'])
+        else:
+            stat_used = config['hyper']['statistic']
+    else:
+        stat_used = 'mean' #default statistics: arithmetic mean over each bin
+
     if isinstance(config['output']['file_name'], list):
         psf_list = config['output']['file_name']
         if len(psf_list) == 0:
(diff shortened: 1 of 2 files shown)


piff/config.py
```diff
@@ -271,23 +289,27 @@ def meanify(config, logger=None):
...
-        average = average.reshape(-1)
-        average[average!=0] /= counts[average!=0]
-        params0[:,i] = average
+        if i in params_fitted:
+            average, u0, v0, bin_target = binned_statistic_2d(coords[:,0], coords[:,1],
+                                                              params[:,i], bins=binning,
+                                                              statistic=stat_used)
+            average = average.T
+            average = average.reshape(-1)
+            Filter = np.isfinite(average).reshape(-1)
```
cpadavis: if I'm understanding this line correctly, then the Filter applied a few lines below will only be for the last param in params_fitted. I don't think Filter is necessarily the same for each parameter. So, I think we want the product of all Filters.

piff/config.py
```diff
@@ -296,6 +318,10 @@ def meanify(config, logger=None):
 
     coords0 = np.array([u0.reshape(-1), v0.reshape(-1)]).T
 
+    # remove any entries with counts == 0
```
cpadavis: I think this comment is not strictly true, in that there are additional other reasons for a box to be non-finite :)

piff/config.py
PFLeget: Yes indeed, I added that in my last commit

piff/config.py
PFLeget: Yes indeed, I specified in the last commit that it was removing bin with non finite value inside

PFLeget:
That's totally fine for me also, and will be useful for the OptAtmo model.

Issue #76, opened 2018-07-24
Use open symbols for plotting negative rho statistic values


jmeyers314:
Using solid and dashed lines to connect positive and negative correlation points helps somewhat, but I think it would be much easier to quickly interpret the rho plots if the symbols used for the  negative points were open (and filled for positive points).

PR #80, opened 2018-12-19
Use open symbols for negative rho values

jmeyers314:
Addresses #76 .

Note, I only verified this works by copy-pasting the `_plot_single` code into a Jupyter notebook and testing there; the Piff test-suite itself doesn't seem to actually produce a plot with negative rho values.


PFLeget:
For me this modification is fine and correspond to what we already discussed. I don't see any problem of pushing that into the master.

PR #83, opened 2019-04-29
REF list galsim like all of the other dependencies

beckermr:
This PR makes sure galsim is properly listed as a runtime environment The current scheme is outside of the pypi standards and is breaking installations from requirements.txt files.


rmjarvis:
Sure.  This stems from before GalSim was pip installable, so we couldn't just include it as a pip requirement.

PR #85, opened 2019-05-02
ensure string ops work for fitsio v1 as well as pre-v1

### LINKED ISSUE #84: make compatible with fitsio v1
fitsio v1 returns strings rather than bytes.

I'll make a PR

esheldon:
closes #84


esheldon:
you're welcome

PR #86, opened 2019-06-28
Galsimify optatmo plus

areshernandez:
Latest Version for Pull Request


rmjarvis:
@areshernandez This PR is not yet ready for review.  Please work on cleaning it up.  (I'm sure you don't actually want me to review 73,000 lines of code!)

There are lots of unnecessary files in here, emacs backup files and "old_atmo_psf_files", which should be removed.

There are also new unit test files, which are just straight copies of other files with most of the calls commented out.  And presumably something changed somewhere to do the new test you wanted.  This needs to be cleaned up.  Probably take the tests that are actually new and relevant and put them in test_optatmo.py.

Also, the unit tests on Travis are failing currently, because you import numba.  AFAICT, this is just used by two functions, which look they could be recast into numpy operations, without the need for explicit loops (and thus not really needing numba).  I think unless you already know that this is really the tall pole for your code, best to just remove the @numba.jit for now and the import, and we can look to make that more efficient later after profiling.

Let me know when this is ready for me to do a real review.

Issue #73, opened 2018-04-13
Make it easier to specify how the PSF profile should be centered on an image


rmjarvis:
Currently `psf.draw(x,y)` centers the PSF profile at the sub-pixel values of the provided x,y positions.

This is sometimes a desirable feature, but not always.  We should make it easier for the user to specify how they want the PSF image centered.  I'm proposing a new `center` keyword argument to the draw command with the following options:
```
center=None   # (default) the current behavior using (x,y) for the sub-pixel position.
center=True     # center the profile at the actual center of the image.
center=pos      # center the profile at some other arbitrary position on the image.
```

esheldon:
works for me

RobertLuptonTheGood:
Why is this needed?  Can't the user just say (int(x), int(y)) to centre in a pixel?  I don't understand the `center=pos` option.  Does it mean calculate the PSF at the specified position, but draw it as if it were at `pos`?

rmjarvis:
FYI, I'll mostly be leveraging the work I did on [this GalSim PR](https://github.com/GalSim-developers/GalSim/pull/1053) where I added a `center` parameter to the `drawImage` command.  This will encompass how the `center=pos` option will work.  The other two options described above will just define two reasonable values for what that pos might be.

PR #88, opened 2019-07-11
Refactor PixelGrid class

rmjarvis:
I've done a big refactoring of the PixelGrid class.

Most of the changes were functionality-preserving, just simplifying some of the implementation, and adding a bunch of inline comments explaining the math involved, since I found a lot of it rather confusing.  I also removed some of the features that we never actually used and didn't really have unit tests to validate.  (E.g. degenerate=False, and mask=False options.)

The main new feature I wanted to get implemented is to give PixelGrid (and thus now all `Model`s) a `getProfile` method, so it could return a GSObject surface brightness profile for a given location.  For PixelGrid this is a `galsim.InterpolatedImage`.

This has the potential to unify all the drawing functionality at a higher level.  So now the base class `Model.drawStar` just gets the profile from the subclass and implements the draw functionality there.  I plan to use this feature to implement #73, but I'll put that off to another PR, since this one is large enough already.

Anyway, the getProfile function was much simpler to implement if I got rid of the flux and centroid constraints that Gary implemented in PixelGrid.  I thought this would not be much more than an implementation detail to handle the constraints at the reflux step, rather than have them automatically satisfied in the matrix solution.  However, it turned out to be an improvement in the quality of the interpolation.

We had been seeing occasional failures of the PixelGrid/BasisPolynomial model in DES solutions when a PSF model was rendered relatively far from any constraining PSF stars.  The edges would sometimes develop a kind of checkerboard pattern.  It turns out this failure mode disappeared after changing how the constraints work.  

Here are the recovered PSF models for a particular Y3 DES image where we had seen the checkerboard failure after (left) my change and before (right).

image

So I think this new implementation is a net improvement.  I tested a small batch (one zone) of Y3 files with the new code and the rho stats were maybe slightly better, but definitely not worse.  I'll run more when CPU time opens up at BNL to get better statistics about this.


esheldon:
Do you have any intuition why this works better?

rmjarvis:
> Do you have any intuition why this works better?

Only a little.  I think it depends a lot on how the star pixels happen to be aligned or not with the model pixels.  If you just have one star, and the model pixels are the same size, but shifted relative to the data by a half pixel, then this checkerboard pattern is not well constrained.  You can get basically the same values in the data pixel centers if you add on a checkerboard pattern, since the contributions from the 4 pixels add and subtract to get back to zero.  There is a version of this checkerboard at any specific pixel offset, although I guess at zero offset, it probably goes away.

With multiple stars at different pixel centers, this is better constrained, but extrapolating far from any PSF exemplars, there can still be a degeneracy direction in the interpolation coefficients that would allow this pattern to show up.

Without the constraints, it seems that there is not much pressure to slide along this degeneracy and have the checkerboard appear.  But when the flux and centroid constraints are added, then they can try to use this degeneracy direction to compensate for some small bit of random noise the might be trying to pull the flux or centroid away from their nominal values.

rmjarvis:
Not sure if anyone feels up to reviewing this.  If so that would be great, but if no one volunteers, I'll plan to merge Friday.  I want to try to get #73 done too and tag a new version (0.3.0) for DESDM to use on Y6.  Brian wants to get that started in August some time I think.

PR #89, opened 2019-08-28
Galsimify optatmo plus cleaning

areshernandez:
Re-created from scratch


rmjarvis:
FYI, I filtered out the large files from the git history, so if you have a local copy of this branch, you can now do
```
git fetch
git rebase
git reflog expire --expire=now --all
git gc --aggressive --prune=now
```
The last two are not really necessary, but they will get your .git directory down to a reasonable size.

rmjarvis:
Some progress: I cleaned up most of the places where there were unnecessary changes to code outside of optatmo.py.  So the file diff now looks a lot cleaner.  I also reformatted lots of lines to 100 character width, so they are more readily viewable on GitHub.

I haven't done any significant substantive changes, but I'll keep hacking on this to continue to clean up the code until it's at a point where I can understand the flow of it.  

I'm pretty sure the overall structure of this needs to change.  For instance I think we want to break out the atmospheric part from this class so we can use GP or other interpolants for that, which doesn't look particularly easy to implement in this structure.

But I can't yet grok the overall flow yet.  There's too much parochial code in here that I don't understand.  So I'll work on updating parts of it to simplify and generalize it, so hopefully it will become more apparent how this can be refactored to get the kind of functionality we want from it.

rmjarvis:
I have a few more questions for anyone who might want to answer:
1. I see a whole lot of code in the fit function involving training stars vs test stars.  There is great care to keep the outlier rejection in the test stars in sync with what was done to the training stars.  But then AFAICT the test stars are just ignored in the end.  Is this intentional?  Or was there supposed to be something done with them that actually impacted the fit somehow?  Or perhaps this is just a relic of some bygone internal-to-the-fit-function testing that has now been excised?
2. I don't much like the names of the new `_cutoff_*` parameters.  But the gist of two of them seem fine.  First, I'd like to change the name `_cutoff_masked_level` -> `max_mask_fraction`.  Not maximum number of masked pixels, but otherwise basically what you already have.  I don't think this cut is required for things like PixelGrid, since it works fine with significantly masked data, but it seems fine as an option which I guess OptAtmo seems to need.
3. Next, I'd like to change `_cutoff_nuisance_level` -> `max_outer_flux` to indicate a maximum fraction of the flux outside the inner portion of the stamp.  Say inner half of the total area or something.  This is relatively close in spirit to what your "nuisance" test is doing.  So let me know if that seems ok based on whatever problems prompted this parameter.
4. The other one `_cutoff_deformed_level` doesn't seem worthwhile.  There is already a `use_partial` option, which defaults to False, so I think that means that no stamps should be "deformed".  Only if you specifically ask to use stars that are partially off the edge of the image should there be any with different x and y sizes.  So I'm inclined to just remove this parameter, but please let me know if there is some reason that you think this parameter is necessary.
5. There is a new test called `test_lmfit_errors` that as far as I can tell is really just a script for someone (Chris or Ares presumably) to try to understand what lmfit reported as its errors and how well they made sense.  It doesn't seem to be actually testing anything about Piff.  So I moved that to the devel directory, rather than keeping it as part of the test suite.  If this was wrong, please let me know and we can move it back (although probably ought to find some way to speed it up, since it takes over a minute now -- fine for a devel script, but not good for a unit test).
6. Maybe related to the above, for me the test `test_atmo_interp_fit` seems to be failing for every star with the error `param.stderr is None`.  It then seems to be ignoring every star with the warning `Warning! Failed to fit atmosphere model for star N. Ignoring star in atmosphere fit`.  We
should probably put in some kind of check that all stars aren't being ignored. (This mechanism was intended to allow for one or two here and there to have a problem and get removed from the star list without causing the whole thing to abort.)  But I wasn't sure whether you were also seeing this on your system?  Or is this some kind of error I accidentally introduced in my intended-to-be-functionality-preserving edits?  Or maybe this is related to the version of lmfit or some other external library, which might be different on my system than yours?

I'll probably have more as I keep going through the code, but that's probably plenty for now.

aaronroodman:
Mike,

Ares will have to answer most of these (and he is away til next Monday).

I thought the test stars were being kept aside for the Rowe-statistics, and also for plotting.

Aaron


On Sep 13, 2019, at 10:00 AM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


I have a few more questions for anyone who might want to answer:

  1.  I see a whole lot of code in the fit function involving training stars vs test stars. There is great care to keep the outlier rejection in the test stars in sync with what was done to the training stars. But then AFAICT the test stars are just ignored in the end. Is this intentional? Or was there supposed to be something done with them that actually impacted the fit somehow? Or perhaps this is just a relic of some bygone internal-to-the-fit-function testing that has now been excised?
  2.  I don't much like the names of the new _cutoff_* parameters. But the gist of two of them seem fine. First, I'd like to change the name _cutoff_masked_level -> max_mask_fraction. Not maximum number of masked pixels, but otherwise basically what you already have. I don't think this cut is required for things like PixelGrid, since it works fine with significantly masked data, but it seems fine as an option which I guess OptAtmo seems to need.
  3.  Next, I'd like to change _cutoff_nuisance_level -> max_outer_flux to indicate a maximum fraction of the flux outside the inner portion of the stamp. Say inner half of the total area or something. This is relatively close in spirit to what your "nuisance" test is doing. So let me know if that seems ok based on whatever problems prompted this parameter.
  4.  The other one _cutoff_deformed_level doesn't seem worthwhile. There is already a use_partial option, which defaults to False, so I think that means that no stamps should be "deformed". Only if you specifically ask to use stars that are partially off the edge of the image should there be any with different x and y sizes. So I'm inclined to just remove this parameter, but please let me know if there is some reason that you think this parameter is necessary.
  5.  There is a new test called test_lmfit_errors that as far as I can tell is really just a script for someone (Chris or Ares presumably) to try to understand what lmfit reported as its errors and how well they made sense. It doesn't seem to be actually testing anything about Piff. So I moved that to the devel directory, rather than keeping it as part of the test suite. If this was wrong, please let me know and we can move it back (although probably ought to find some way to speed it up, since it takes over a minute now -- fine for a devel script, but not good for a unit test).
  6.  Maybe related to the above, for me the test test_atmo_interp_fit seems to be failing for every star with the error param.stderr is None. It then seems to be ignoring every star with the warning Warning! Failed to fit atmosphere model for star N. Ignoring star in atmosphere fit. We
should probably put in some kind of check that all stars aren't being ignored. (This mechanism was intended to allow for one or two here and there to have a problem and get removed from the star list without causing the whole thing to abort.) But I wasn't sure whether you were also seeing this on your system? Or is this some kind of error I accidentally introduced in my intended-to-be-functionality-preserving edits? Or maybe this is related to the version of lmfit or some other external library, which might be different on my system than yours?

I'll probably have more as I keep going through the code, but that's probably plenty for now.

rmjarvis:
Another question:

Is the `vary_optics=True` options for `fit_model` ever useful?  It seems to only be used in the unit tests as an explicit specific test of this option.  None of the presumable more "real world" tests use it, and all the calls to `fit_model` have `vary_optics=False`.  So is this a relic of some past attempts that didn't work?  Or is there a use case that we need, which isn't being exercised at the moment in the tests, which would set this to `True`?

rmjarvis:
Update: I converted reflux and fit_model to use scipy rather than lmfit.  The code there is a fair bit simpler now, and the test functions that run it are around 10-20% faster, so that's nice.

I haven't done the optics fits yet.  I wanted to start with the pixel one, but that seems to be broken currently.  (There is no unit test of it, but when I enable mode='pixel' in test_fit, it fails.)  So I'll work on that next to see if I can get that working with the current structure before converting it over to scipy.

rmjarvis:
Update: I converted all the fits to use scipy.optimize.least_squares directly rather than lmfit.minimize.  I also fixed the pixel mode to work correctly (which it hadn't been), and it's only a little bit slower than the shape mode: `test_fit` takes 38 seconds for `pixel` and 34 seconds for `shape`.

The next bit I want to ask about is some of the cuts of the `stars` list you have in the `fit` function.  They are mostly based on the measured shapes, but it seems like they are probably too strict currently, since they cut the stars list down from 117 stars to only 17 in test_fit.  Considering that the stars in this test don't have any data-related problems (they are simulated in the test), that seems rather extreme.  I can see having outlier rejection to remove stars that are maybe too noisy or too close to a neighbor, but removing 85% of a set of simulated stars with none of these problems doesn't seem great.  So I'd like some justification of why you thought these cuts were necessary.

I also don't really like this kind of outlier rejection to be hard-coded within the OptAtmoPSF class.  The outlier rejection should really be encapsulated in an Outlier step.  We can create an Outlier type based on the moments as an alternative to the current Chisq outlier type.  Alternatively, we can add an option to make these cuts initially on the input, before even getting to the OptAtmo fitting.

The one the seemed potentially ok to keep in OptAtmo is the limit to the 500 brightest stars.  That's not per se a rejection of bad stars, just that we don't really need more than this many, so save some time by limiting to just these.

areshernandez:
Hi Mike,

I finished reading your changes that were made while I was on vacation. I will add a full response including all the answers to your questions and concerns I have shortly.

areshernandez:
This is the full list of concerns/problems I see with the changes that have been made and their explanations. Most problematic changes appear to have occurred due to misconceptions about what code was obsolete/changeable and what was not, most notably the very important vary_optics and vary_shape parameters and all the changes that were built on top of eliminating them. I did not check github when you asked questions about whether or not you could make these changes because I was on vacation. These are not in order of importance but rather (roughly) in chronological order with older changes coming first. A few of these concerns are actually me saying a certain change was fine but I thought it would be important to mention that for some reason.

1. Removing the initial starting guess for flux in star_stats.py. I remember Chris added this initial guess to several parts of the code that fit the flux because if this guess is not used the fit fails very badly most of the time.

2. Moving test_lmfit_errors() to the new devel directory you made is perfectly ok. I do not know who made this test (presumably, Chris, I guess).

3. Your suspicions about the fit_atmosphere() function are correct. It is no longer in use and has been replaced by code in the PIFF fitting pipeline. With some help from Josh, I was able to change this code so that you can run a fit that does optics and individual star fitting for the atmosphere and then stops. You can then run the fit for just the GP atmospheric interpolation afterward (or any other interpolation). You can also do both at once. I have had conversations with Aaron about this, and I may very well remove fit_atmosphere() from the code altogether in the future as well as many functions in optical_model.py.

4. I think Aaron covered this, but the random forest fit is very important. The code is very slow and doing an initial fit to get close to the correct param values could help a lot. Leaving the random forest optional for now is fine, however. If Anna (the grad student that Aaron mentioned) manages to find a way to get it to work such that we not only get significant speed up but also higher fit quality I would prefer it be reverted back to not being optional I think keeping it optional for the time being should be fine.

5. I'm not 100% certain but I think you had a concern with including opt_size, opt_g1, opt_g2, and opt_L0 (if vonkarman) in the optical fit. This cannot be changed. This is a core design choice that was made very early and I think I remember Chris telling me at some point that there is no way to run a "pure" optical fit without encountering very bad fits if you exclude opt_size, opt_g1, and opt_g2. These are the average of the size (inversely proportional to r0, the Fried parameter), g1, and g2 values across the focal plane.

6. Outlier rejection for test stars very carefully matching the train stars is very important. This is because the test stars are used extensively for rho statistics and many other graphs in the PIFF fitting pipeline.

7. Regarding the names of the _cutoff parameters for star cuts, I encountered a strange error when I first created them (I can't remember where) unless I started their names with an underscore. 

8. I disagree with changing the masked star cuts to work with masked_fraction rather than number of pixels masked. These cuts should be used if you want to cut based on having a small number of pixels or even just one pixel masked depending on how strict you want to be (if you want to use this cut). I'm open to just changing the name, of course.

9. Likewise, I'm fine with renaming the nuisance star cut so long as it is not actually changed.

10. In my experience, I have encountered "deformed" postage stamps from stars on the edges, so I want the deformed star cuts to stay.

11. I (and also Josh) have also encountered the "param.stderr is None" issue myself although not (I think) to the extent of losing all my stars, at least not currently. I do not know the cause currently.

12. horner2d() fails for its most recent iteration in galsim because it does not work with complex numbers. I talked to Josh about this and he said he would eventually either write some code that would allow getParamsList() in PIFF to work without complex numbers or change horner2d() in galsim, so that it starts working with complex numbers again. In the meantime, the horner2d_from_galsim1() functions are necessary.

13. This is very critical. Removing vary_optics and vary_shape is not possible. Both are used extensively for various code written by both me and Chris when I want to do single star fits and (at least vary_shape) is used in very many cases in parts of the PIFF fitting pipeline written by both me and Chris. This would be very time consuming to fix as the amount of code that has been written outside of PIFF that is used very often by me and others such as Pierre-Francois that depends on these two variables is very large. Several of the following concerns I have coming up stem from changes that were made building off of the elimination of these two variables. I have also strongly considered adding uses of both of these variables into PIFF itself and even into the fit() function, although that will likely be a change done in the future.

14. Due to problem 13, removing the if/else statement in in _fit_model_residual() for when vary_optics is True and optical_profile is None is not possible.

15. Hardcoding things like self.n_params_atmosphere, etc. could be potentially very problematic. Currently, (I think) the code is in its optimal form with having an opt_L0 and no atmo_L0 as I've done many tests with many combinations of opt_L0 and/or atmo_L0 and found that using only opt_L0 is best because it is too difficult to fit atmo_L0 precisely. Although, I reached this conclusion awhile ago and it probably will not change this code has gone through several iterations of using atmo_L0 or opt_L0. In the (unlikely) event that putting atmo_L0 back in becomes a good idea at some point in the future having all these hardcoded values would make this task difficult.

16. Due to problem 13, separating out the optical params in fit_model and labelling them as "const_params" is not possible as I want those params to be able to vary if vary_optics=True. Several of the following problems are due to changes that were build off of this change.

17. I did not create the "no estimated errorbars" check and I do not know what Chris uses it for. I am concerned that it may be important. Also, I created the estimated_errorbars_not_required variable because there is some code in the PIFF fitting pipeline that I created that I know for sure does not require estimated errorbars to do what I wanted to do. If the "no estimated errorbars" check is left in, estimated_errorbars_not_required should be left in as well.

18. Changing fit_model() to combine opt_size & atmo_size together in the fit goes against a fundamental design convention of the code to always be very careful and keep the two separate. This is especially important due to the plethora of things vaguely just called "size" in the code that I have found has strongly confused everyone who has used/worked with the code except Chris himself. I will talk about a problem similar to this one in another problem coming up.

19. In my experience, I am very concerned about the replacement of size and flux with log(size) and log(flux) in many changes to the code. I have encountered many problems making changes with the fitting with these parameters. They are very finicky and I am very concerned that fitting their logs rather than them themselves could change things a lot.

20. I am also very concerned about replacing drawProfile() with drawImage() in _fit_model_residual(). This too is finicky and I've encountered many issues many times due to minute differences between using drawStar() vs drawProfile vs drawImage(). This doesn't always happen, but to be safe, this should not be changed.

21. It is not possible to replace lmfit in fit_model() with scipy as I need to be able to pick and choose parameters to float and retrieve the fit_results as well. 

22. It is not possible to rename the "kolmogorov_kwargs" as just "kwargs" as, in a similar fashion to problem 16, there are large amount of things vaguely referred to as just "kwargs" in this code and, in my experience, this sort of thing has generated confusion. I am open to renaming this something like "atmo_kwargs" or something like that but just "kwargs" would defenitely generate a lot of confusion.

23. It is not possible to remove initial guesses for fit_model(). They are used many times in the PIFF fitting pipeline with vary_shape=True for doing speedier individual star fits. I can't remember if I have tested exactly by how much this is faster but it would not be a simple test as this is something currently external to PIFF proper that could not be easily tested by a unit test.

24. It is not possible to remove pull cuts for pixel-chisq fitting. These cuts are not just used in the optical fit but are then used by the PIFF fitting pipeline aftwerward for various purposes regardless of whether using pixel-chisq or shape-chisq fitting.

25. In similar fashion to problem 16, it impossible to remove many of the comments I have made explaining how the size parameter is inversely proportional to the Fried parameter. In my experience, every person that has worked with/used my code except Chris himself has suffered from enormous confusion due to the very large number of things in the code simply called "size." I have sometimes even found difficulty in explaining the code to other people specifically due to this very issue. I find it essential to extensively comment the code to be as specific as possible whenever "size" appears so that people never get mixed up when seeing various things such as "opt_size," "atmo_size," "size," "e0," "r0," etc.

26. Going beyond what was mentioned in problem 25, many comments in general that were removed/shortened need to be there in their original form, even if long. In my experience, this code is something that is extremely confusing to the un-initiated and I have encountered a lot of trouble explaining it to many people several times. I spent a very large amount of time adding comments to the code (in its original form, it was almost barren of comments and what comments were in there were vague, confusing, and so brief they actually generated more confusion). This code is special in the amount of confusion that I have seen it deliver to various people, so the large amount of comments put in there all have origins in clearing up various specific pieces of confusion that I know people have encountered.

27. It is not possible to remove the brute_force, shapes, number of fit steps, lmfit, or random forest caches from fit_size(). All of these were added  to make  fit_size() work best. Fit_size() is essential and needs to be very precise to get a good initial value of the opt_size parameter, which is the most important one and is far off without fit_size() being used and being used to the most precision.

28. Aaron and I both discussed the higher order reference wavefront and we concluded that it definitely should be in the unit tests as we want it tested. It should not be removed from the unit tests.

29. In similar fashion to problem 21, it is not possible to replace lmfit in any of the fit_optics functions with scipy, except in the pixel-chisq mode case.

30. In general, all star cuts are necessary and were added due to large problems encountered in various real fits (since fixed). I have looked into this and the exposure used in the unit tests is special in the sheer number of stars that get cut. This does not happen for the overwhelming majority of exposures I use in real fits. I will discuss the reasons for all the star cuts I can remember off the top of my head right now (I may be forgetting a couple). The max shapes cuts are needed because such stars cause problems in real fits. They are very hard to fit and create detectable problems when analyzing the fits with the PIFF fitting pipeline. SNR cuts and MAD cuts on shapes are in for similar reasons. Pull cuts are needed for similar reasons but are also then used by the pipeline for many other tasks after the optical fit is done too. Nuisance star cuts (excessive outer flux cuts) are there because although only a small fraction of stars fall into this cut for real fits on exposures, fitting stars with an extra nuisance star in the outer edge of the postage stamp is disastrous for the fit. Something similar prompted the masked star cut. The deformed star caused errors during some unit tests I created that were only fixed when I added them. There are very few that fall into this cut. Some stars fail to have their shape measured and are cut since shapes are essential to the code.

31. Switching the cuts to use the Outlier class would be very time consuming. I do not know how to use this class and there are many cuts that I have created independently of the Outlier class because I did not know it existed when I made them. Beyond what you see in PIFF proper, the PIFF fitting pipeline uses many of the cuts for various purposes after the optical fit is done in many places. These cuts are used in many places throughout many pieces of code I have created for the PIFF fitting pipeline. In principle, I am not 100% against absorbing some of my star cuts into the Outlier class at some point in the future but I definitely believe if this done, it should be much after taking care of everything else as this would require a lot of re-coding and testing that nothing gets broken.

32. This list should have (at least almost) all problems introduced by the recent changes. Beyond this, however, I am very concerned that some changes not in this list could create many overlapping bugs not caught by the unit tests. In my experience, the code is extremely finicky and many even seemingly tiny changes that I have made to this code have completely broken and caused large errors in doing a real fit that is completely undetected by the unit tests. In the event that this code, only with the changes not in this list, passes the unit tests, my ability to run real fits may still break. If this happens and it seems to be an intractable problem to get things running again, it may be necessary to revert back more changes than are in this list.

areshernandez:
Hi Mike,

I have developed a To-Do-List for this Pull Request and a much shorter list of the potential problems with the recent changes made to the code (focusing only on the most important ones and explaining them better). Most of the To-Do-List is creating Unit tests to test if many of these potential problems in my list actually worsen the fit quality.




To Do List:

1. Restore higher-order reference wavefront for test_fit(), and test both point-cloud (data) and grid (Zemax) reference wavefronts.  Note that the latter are typically using for z12,13 and z16 and higher. Also, make sure the higher order wavefront is tested with z12, z13, z16, etc.

Status: Done

2. Finish making all unit tests.

Status: Done but currently checking if unit test #4 actually discovered a real bug or not.

3. Attempt a fit on one exposure with the PIFF fitting pipeline with Mike’s version of the code and see if it runs and gives consistent results with previous code

Status: Currently makes it to the first instance of fit_model(), where it stops due to not having the vary_shape option.



New Unit tests:


1. Do an optical fit, using results of said optical fit to then do an atmospheric fit with one star with the method fit_model() with vary_shape=True. See if the optical and atmospheric parameters found both match the ones used to generate the star or not. Will put in devel area.

Status: Passed for both my and Mike’s version of the code.

2. Do a flux and centroid fit for one star, using results of an optical fit using the method  fit_model() with vary_shape=False, and see if a star residual plot can be made and if the pixel values are similar for the model to the data. Will put in devel area.

Status: Passed for my version of the code. Cannot test with Mike’s version of the code yet without vary_shape.


3. Perform single-stars fits to atmo_ parameters for many stars, and check that fits converge to atmo_parameters used in the generation of the stars. Will put in devel area.

Status: Passed for both my and Mike’s version of the code.

4. Run fit_size(), testing that the value of the opt_size parameter matches what was supplied to create the stars used in the test. Will put in devel area.

Status: Passed to tolerance for my version of the code. Did not pass to tolerance for Mike’s version of the code (currently looking at this to see if this was a real bug discovered or some mistake made when copying the unit test I made from my version of the code to Mike’s version).

5. Full exposure fit (all CCDs) to nominal set of floating parameters.  Will put in devel area.

Status: Not yet started.



Questions/Concerns:

1. vary_optics and vary_shape arguments for the fit_model() function have functionality not currently in unit tests.  New unit tests #1 and #2 above test this functionality. This functionality is often used for exploratory studies and is always used in the PIFF fitting pipeline for all real fits we do.

2. Some removed comments in the code were aimed at conceptual explanations of the code. Some of these include ones carefully specifying/describing the various variables related to “size,” which I’ve seen confuse people.

3. Replacing long variable names with hardcoded numbers is likely to generate confusion due to being more ambiguous. We propose replacing them with much shorter but still instructive variable names. On a similar note, I propose renaming kolmogorov_kwargs as “atmo_kwargs”, so it is not confusing for the Vonkarman case but is less ambiguous than “kwargs.”

4. The use of lmfit should be restored. I’ve spoken with Josh and he has helped me craft and organize my strongest set of arguments for why this should be:

  A. We use the functionality in lmfit to change whether parameters are floating or fixed, functionality needed in unit test #2 above (and elsewhere).  It isn’t obvious if scipy.optimize.leastsq has this functionality.

  B. fit_model() creates an lmfit results object. This results object has good functionality, putting the params into a clear, unambiguous dictionary. As Josh mentioned,  “access to parameters by name instead of order means you don’t have to remember what order you used for the parameters.” I have these perfectly memorized but that is only because I’ve been working on it for so long.


  C. Josh also mentioned to me that “adding bounds is easy in lmfit, where it internally uses some kind of sigmoid transform to limit the range of parameters.” I use and change lmfit’s bounds numerous times when doing different fits. So having an easy way to do this is very important.

areshernandez:
Hi Mike,

Also, I wanted to add that I will git push these unit tests and other small tweaks I made later today. I did not make significant changes regarding things from the list. For example, I have not re-introduced vary_optics or vary_shape. I mainly changed just what I had to so I could test what I wanted to test.

rmjarvis:
> Do a flux and centroid fit for one star, using results of an optical fit using the method fit_model() with vary_shape=False, and see if a star residual plot can be made and if the pixel values are similar for the model to the data. Will put in devel area.

This functionality is reflux now.  Not fit_model.

rmjarvis:
> The use of lmfit should be restored. 

Consider this a non starter for now.  The code is vastly more maintainable without lmfit.  I'm not willing to restore it without a **very** compelling use case where scipy simply cannot do the job.  Not where you and Josh merely find the API more convenient.

rmjarvis:
> Status: Currently makes it to the first instance of fit_model(), where it stops due to not having the vary_shape option.

In a test like this, you shouldn't be calling fit_model and other such things by hand.  You should be calling `piffify` with a config file.  That's the usage that we are trying to enable.  End users will not be calling the guts of the code themselves.  So the new code will not fail for not having a `vary_shape` option, since it never tries to do that.

rmjarvis:
Latest updates are mostly code speedups.  But also a big refactor of all the various `hsm_*` functions into a single function called `calculate_moments`.  This has the nice side effect of getting the code coverage for this PR patch up to nearly 99%, which means it passes the requirements.

The new moments calculation are not quite the same as what you used to have.  I made the following changes:
1. sigma2_normalization had an error in that the weight and kernel were squared twice rather than just once.  This seems to be clearly wrong.
2. The various `sigma2_*_u0` and `_v0` quantities squared `weight * kernel * data / normalization`, which doesn't make any sense to me.  I'm pretty sure these factors should not be squared.
3. Points 1 and 2 made a difference to the overall values, so I had to adjust the fudge factors from > 2 to around 0.8 to make `test_snr_and_shapes` to continue to pass.  These changes are in commit ee95e38345098b2405e5dbda0520d10ef8e5cbb5 if you want to see them directly.
4. There were a couple other typos that I fixed, which made essentially no noticeable change in anything that mattered.  [One was a `*2` which should obviously have been `**2`](https://github.com/rmjarvis/Piff/blob/ee95e38345098b2405e5dbda0520d10ef8e5cbb5/piff/util.py#L1262) and [the other was a `du**2 - dv**2` which should have been `du**2 + dv**2`](https://github.com/rmjarvis/Piff/blob/ee95e38345098b2405e5dbda0520d10ef8e5cbb5/piff/util.py#L1260).  I didn't single these out into their own commit, since it didn't notice them until much later.  They are fixed as part of commit c58f82df5d611654843910bb8d067210e8a2593b.
5.  I couldn't find any justification for the extra factor of 2 you added to all the moments after the first 3.  So I removed it in `calculate_moments`, so they all return a coherent set of moment definitions.  However, to get back to your definitions, the functions in the OptAtmo class that call `calculate_moments` put the factor of 2 back in.  See the `if True` blocks, e.g. [here](https://github.com/rmjarvis/Piff/pull/89/files#diff-3f888c72d9461d4f6a3739bcc9e9fe90R1498)
6. Similarly, I found your definition of the "flux" moment to be a bit weird.  For a Gaussian profile, your code returned a "flux" of flux**2 pixel_scale**2 * mean(weight) / 4 pi sigma**2.  The new code simplifies this somewhat to just flux / 4 pi sigma**2, which seems better, since it is independent of the image's pixel scale or weight value, both of which seem like they shouldn't directly change the moment.  Anyway, this again is restored to your original value in those `if True` blocks.
7. A more minor difference is in the centroid moments.  Your code includes the hsm centroid value as part of the measured moment.  This is probably not the best definition for this, since the location of the profile in the image is arbitrary, and not generally exactly at the image center (the "(0,0)" position for hsm).  The new code returns the first moments *relative* to the hsm center, which is more fundamental to the profile shape.  E.g. for a Gaussian, this is always very close to (0,0) regardless of the sub-pixel center of the profile.  But other aberrated profiles give non-zero results here, still independent of exactly where on the image the profile is drawn.  So I think this is better, but again your original definition is restored by those `if True` blocks.

I suspect we can eventually remove the `if True` blocks that account for changes 5,6,7.  But it would require you retraining the random forest with the new moments.  Probably a good idea anyway, since changes 1,2,3,4, which are actual bugs, presumably already invalidate the current RF solutions.

areshernandez:
Just finished git pushing 4 unit tests for the devel area as well as some small fixes for the higher order RW. I have found out that test_star_residuals() broke after the most recent git push of changes. This is due to some change in reflux(), as using the version of reflux() before the push works (I left that old version in commented out so this is easy to check). Test_size_fit() also fails. After enough reviewing and re-testing, I have convinced myself that this is a real bug of some kind. The fit to opt_size is not to within tolerance and is further off with the most recent git push.

rmjarvis:
Thanks Ares.  

The first thing you noticed was indeed a bug.  The new reflux didn't work well if the initial guess is many orders of magnitude wrong (1 vs 1.e6 in this case).  So I changed it to use the sum of pixels as an initial guess if the current value is 1 (i.e. unknown yet).

The second thing you mentioned, I'm not seeing.  test_size_fit works for me.  Although, I didn't try running it until I fixed reflux, so maybe that was the same bug showing up again.

I was also finding that `test_optics_and_test_fit_model` was failing.  It took me a while to track this down, but it seemed to be because of the changes to the error estimates for the moments.  When I put back in some of the known-to-be-wrong formulae that had been in the old code, it worked.  This clearly isn't the right solution, since those formulae are definitely not right.  But something about the way they were wrong helps the fitter work better.  

Of course, the new estimates were still not that great (just more accurate than previously), so it's possible they just need to be improved further.  Maybe actually good error estimates will work well.  

But it's also possible that we need to add some weighting based on the flux or something, since the nature of the error in the old estimates was that they were too large by a factor of the flux.  (Not precisely, but that's the zeroth order effect.)  I tried a few iterations of that and couldn't find something that worked other than the old way of estimating the errors, but it definitely doesn't seem like the right solution.

Anyway, for now, I've put back in the old estimates, and this test passes.

rmjarvis:
Also, please revert the recent change to decam_wavefront.  This would be another new feature to have the reference wavefront go beyond z=11.  Let's stick to the feature set that's in here now and stop adding new features.  If you want to add this feature later, let's please make that a different pull request where we can review that on its own.

aaronroodman:
Mike

The initial pull request had the reference wavefront going up to z=37, and having these high order terms was needed to do a better job on the detailed PSF shape, ie 3rd and 4th order moments.

My conclusion is that we need these terms

Aaron

Sent from my iPhone

On Oct 10, 2019, at 6:47 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


Also, please revert the recent change to decam_wavefront. This would be another new feature to have the reference wavefront go beyond z=11. Let's stick to the feature set that's in here now and stop adding new features. If you want to add this feature later, let's please make that a different pull request where we can review that on its own.

aaronroodman:
And... basic operation and speed can be tested without these - but the details of the optical fit will be off

Sent from my iPhone

On Oct 10, 2019, at 6:47 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


Also, please revert the recent change to decam_wavefront. This would be another new feature to have the reference wavefront go beyond z=11. Let's stick to the feature set that's in here now and stop adding new features. If you want to add this feature later, let's please make that a different pull request where we can review that on its own.

rmjarvis:
> The initial pull request had the reference wavefront going up to z=37, and having these high order terms was needed to do a better job on the detailed PSF shape, ie 3rd and 4th order moments.

No. That's a different thing.  That's the `higher_order_reference_wavefront_zernikes_list` which used to be from 12-37.  Ares recently added a feature to allow `reference_wavefront_zernikes_list` to be from 4-11 plus 14 and 15.

aaronroodman:
Ah ok! - that fix can indeed wait

Sent from my iPhone

On Oct 10, 2019, at 7:06 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


The initial pull request had the reference wavefront going up to z=37, and having these high order terms was needed to do a better job on the detailed PSF shape, ie 3rd and 4th order moments.

No. That's a different thing. That's the higher_order_reference_wavefront_zernikes_list which used to be from 12-37. Ares recently added a feature to allow reference_wavefront_zernikes_list to be from 4-11 plus 14 and 15.

areshernandez:
Hi Mike,

I removed the ability to pick and choose which zernikes you want for the higher order reference wavefront. I reverted back to the old harcoding.

rmjarvis:
FYI, I fixed the Travis installation problem with using GalSim master branch, so it's running the tests again now.  However, there are lots of new failures from the recent commits.

I also made a branch called simply `optatmo` that has a version from here a while back that was passing all the tests. (Specifically ade8d23f4d8def92bf4486083c4c24f9ba2438fc)  We discussed doing this some time ago, so now this is finally done.

I'm inclined to close this PR now and let you do sandbox development on this branch (or rename if you want, since the name is pretty cumbersome).  Then if you think you have something to merge into `optatmo`, you can reopen a new PR at that point, and we can take a look.  But it doesn't seem helpful to leave this PR open as you are working in this kind of exploratory development mode.

You can also feel free to take coherent features here that you think are well vetted to merge into `master`.  For these, make a new branch off of master, copy over the relevant code bits (including tests!) from this branch, and make a PR for them.  If we do these in smaller bite-size chunks than this PR, that would be better for doing a useful code review of each feature set.

rmjarvis:
Closing the PR.  @eacharles  is working on making more bit-sized PRs of this functionality, which will be reviewed separately.

PR #91, opened 2019-09-20
Remove dependency on lmfit

rmjarvis:
I think some of the issues I'm having with Ares's pull request stem from some API changes that lmfit made between 0.8 and 0.9.  I suspect Chris or Ares originally wrote the code to the old API specification and the changes broke various aspects of the code.

Rather than update the code to use the new API (especially since I'm not really familiar with either the old or the new one -- I don't ever use lmfit myself), I'd like to switch everything over to scipy.optimize.

So to get that conversation started, and also to help myself understand what the differences are between the two, I first decided to switch over the other uses of lmfit that we have in Piff.

The only changes to the unit tests required were some checks to only 8 decimal places of accuracy rather than 14.  The latter seems rather overkill, and I guess must stem from some more precise defaults in the lmfit package than scipy's default ftol=1.e-8.

Does anyone have any objections to this?  If not, I'm happy to do the work to switch the new optatmo stuff to use scipy instead of lmfit.

### DIFF
--- a/piff/gsobject_model.py
+++ b/piff/gsobject_model.py
@@ -106,20 +106,19 @@ class GSObjectModel(Model):
             du, dv, scale, g1, g2 = params
         return self.gsobj.dilate(scale).shear(g1=g1, g2=g2).shift(du, dv)
 
-    def _lmfit_resid(self, lmparams, star):
-        """Residual function to use with lmfit.  Essentially `chi` from `chisq`, but not summed
-        over pixels yet.
+    def _resid(self, params, star):
+        """Residual function to use with least_squares.
 
-        :param lmparams:  An lmfit.Parameters() instance.  The model.
-        :param star:    A Star instance.  The data.
+        Essentially `chi` from `chisq`, but not summed over pixels yet.
+
+        :param params:      A numpy array of model parameters.
+        :param star:        A Star instance.
 
         :returns: `chi` as a flattened numpy array.
         """
         import galsim
         image, weight, image_pos = star.data.getImage()
-        flux, du, dv, scale, g1, g2 = lmparams.valuesdict().values()
-        # Fit du and dv regardless of force_model_center.  The difference is whether the fit
-        # value is recorded (force_model_center=False) or discarded (force_model_center=True).
+        flux, du, dv, scale, g1, g2 = params
 
         # We shear/dilate/shift the profile as follows.
         #    prof = self.gsobj.dilate(scale).shear(g1=g1, g2=g2).shift(du, dv) * flux
@@ -128,12 +127,8 @@ class GSObjectModel(Model):
         jac = galsim._Shear(g1 + 1j*g2).getMatrix()
         jac[:,:] *= scale
         flux /= scale**2
-        if galsim.__version__ >= '2.0':
-            prof = galsim._Transform(self.gsobj, jac.ravel(), offset=galsim.PositionD(du,dv),
-                                     flux_ratio=flux)
-        else:
-            prof = galsim._Transform(self.gsobj, *jac.ravel(), offset=galsim.PositionD(du,dv),
-                                     flux_ratio=flux)
+        prof = galsim._Transform(self.gsobj, jac.ravel(), offset=galsim.PositionD(du,dv),
+                                 flux_ratio=flux)
 
         # Equivalent to galsim.Image(image, dtype=float), but without the sanity checks.
         model_image = galsim._Image(np.empty_like(image.array, dtype=float),
@@ -146,18 +141,13 @@ class GSObjectModel(Model):
         model_image.array[:,:] *= np.sqrt(weight.array)
         return model_image.array.ravel()
 
-    def _lmfit_params(self, star, vary_params=True, vary_flux=True, vary_center=True):
-        """Generate an lmfit.Parameters() instance from arguments.
+    def _get_params(self, star):
+        """Generate an array of model parameters.
 
         :param star:         A Star from which to initialize parameter values.
-        :param vary_params:  Allow non-flux and non-center params to vary?
-        :param vary_flux:    Allow flux to vary?
-        :param vary_center:  Allow center to vary?
 
-        :returns: lmfit.Parameters() instance.
+        :returns: a numpy array
         """
-        import lmfit
-
         # Get initial parameter values.  Either use values currently in star.fit, or if those are
         # absent, run HSM to get initial values.
         if star.fit.params is None:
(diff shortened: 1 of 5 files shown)


aaronroodman:
Mike

I don’t think we have ever done a dedicated study of optimization method for the various fits - so I’m not opposed in principle to changing optimizers.

On the other hand scipy.optimize is not a simple replacement for LMfit, since LMfit uses a vector of inputs instead of a scalar chi2 - as you probably know.

Or are suggesting using optimize.leastsq ?  I think LMfit has a lot of additional functionality that leastsq doesn’t

I believe Ares will be back Monday and could work to rationalize the current use of LMfit as well

Aaron







Sent from my iPhone

On Sep 20, 2019, at 2:50 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


I think some of the issues I'm having with Ares's pull request stem from some API changes that lmfit made between 0.8 and 0.9. I suspect Chris or Ares originally wrote the code to the old API specification and the changes broke various aspects of the code.

Rather than update the code to use the new API (especially since I'm not really familiar with either the old or the new one -- I don't ever use lmfit myself), I'd like to switch everything over to scipy.optimize.

So to get that conversation started, and also to help myself understand what the differences are between the two, I first decided to switch over the other uses of lmfit that we have in Piff.

The only changes to the unit tests required were some checks to only 8 decimal places of accuracy rather than 14. The latter seems rather overkill, and I guess much stem from some more precise defaults in the lmfit package than scipy's default ftol=1.e-8.

Does anyone have any objections to this? If not, I'm happy to do the work to switch the new optatmo stuff to use scipy instead of lmfit.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/rmjarvis/Piff/pull/91

Commit Summary

  *   Don't use lmfit for gsobject_model.reflux
  *   Don't use lmfit in star_stats
  *   Don't use lmfit in gsobject_model least squares fitting
  *   Remove lmfit dependency

File Changes

  *   M piff/gsobject_model.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-0> (138)
  *   M piff/star_stats.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-1> (37)
  *   M requirements.txt<https://github.com/rmjarvis/Piff/pull/91/files#diff-2> (1)
  *   M setup.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-3> (2)
  *   M tests/test_gsobject_model.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-4> (4)

Patch Links:

  *   https://github.com/rmjarvis/Piff/pull/91.patch
  *   https://github.com/rmjarvis/Piff/pull/91.diff

aaronroodman:
Also the LMfit release notes for v.9 describe the non backwards compatible changes they made and how to update - if that’s the issue it looks very easy to fix



Sent from my iPhone

On Sep 20, 2019, at 2:50 PM, Mike Jarvis <notifications@github.com<mailto:notifications@github.com>> wrote:


I think some of the issues I'm having with Ares's pull request stem from some API changes that lmfit made between 0.8 and 0.9. I suspect Chris or Ares originally wrote the code to the old API specification and the changes broke various aspects of the code.

Rather than update the code to use the new API (especially since I'm not really familiar with either the old or the new one -- I don't ever use lmfit myself), I'd like to switch everything over to scipy.optimize.

So to get that conversation started, and also to help myself understand what the differences are between the two, I first decided to switch over the other uses of lmfit that we have in Piff.

The only changes to the unit tests required were some checks to only 8 decimal places of accuracy rather than 14. The latter seems rather overkill, and I guess much stem from some more precise defaults in the lmfit package than scipy's default ftol=1.e-8.

Does anyone have any objections to this? If not, I'm happy to do the work to switch the new optatmo stuff to use scipy instead of lmfit.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/rmjarvis/Piff/pull/91

Commit Summary

  *   Don't use lmfit for gsobject_model.reflux
  *   Don't use lmfit in star_stats
  *   Don't use lmfit in gsobject_model least squares fitting
  *   Remove lmfit dependency

File Changes

  *   M piff/gsobject_model.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-0> (138)
  *   M piff/star_stats.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-1> (37)
  *   M requirements.txt<https://github.com/rmjarvis/Piff/pull/91/files#diff-2> (1)
  *   M setup.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-3> (2)
  *   M tests/test_gsobject_model.py<https://github.com/rmjarvis/Piff/pull/91/files#diff-4> (4)

Patch Links:

  *   https://github.com/rmjarvis/Piff/pull/91.patch
  *   https://github.com/rmjarvis/Piff/pull/91.diff

rmjarvis:
It seemed like [`scipy.optimize.least_squares`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#scipy.optimize.least_squares) was pretty close to a drop-in replacement for `lmfit.minimize`.  At least as it was used in the other places we were using it.

But if you think there are important features that lmfit has (which we are using), but scipy doesn't, please do let me know.

jmeyers314:
I'm pretty sure `lmfit` just wraps `scipy.optimize`, e.g., https://github.com/lmfit/lmfit-py/blob/d7b408baa64821f4356cff9c32a199bc53becd93/lmfit/minimizer.py#L1438.

There are some potentially nice features of using the wrapper; like the ability to fix and unfix different parameters and fit them each in sequence.  I think that would probably look less elegant with `spicy.optimize`.  I don't know if we're currently using or would benefit from such features in Piff though.

rmjarvis:
If it's just wrapper around scipy, then I would prefer to get rid of it.  
1. I actually find the API to be more cumbersome than directly using scipy.optimize.least_squares, making lmfit.ParameterList objects rather than just using numpy arrays.  It sounds like the two of you might disagree with me about this, but I thought the code was easier to follow here after switching to scipy.
2. The lmfit package is still in `0.x` versioning, and they apparently aren't yet at the point where they care about being backwards compatible.  Breaking user code on a minor version upgrade without a previous deprecation stage is not something I particularly want to deal with.  scipy's API is much more likely to remain consistent and not break things if users of Piff upgrade to a new scipy version.
3. The lmfit front end does seem to add a not-completely-negligible overhead.  When I switched the gsobject implementation from lmfit to scipy, the time to run test_gsobject_model dropped from 27 seconds to 24 seconds.  So 10% faster overall, of which only a fraction (although possible a majority) is spent in the relevant functions.  So it seems that lmfit adds an overhead of at least 10% for the way that we were calling it in that class.  Not crazy, but not negligible either.  If there's not a good reason to incur that overhead, I'd rather not.
4. While we have been using the feature that Josh pointed out of turning on and off parameters to be fit, I find that it makes the code significantly harder to follow, since whether things vary can be set in other places than where you have to know how many parameters are being varied.  I'd rather restructure that code to call least_squares with just the parameters being varied, rather than kind of hide this behavior in the lmfit_params object.  This may be another point of difference between me and the two of you.  I suppose YMMV.

Issue #82, opened 2019-02-22
Change the order of operations in loading images to reduce memory footprint


rmjarvis:
@PFLeget  had trouble starting up Piff running on a whole LSST exposure, since it currently loads all the images into memory and then extracts all the stars from them.  There's no need to do it this way.  It would be much more memory efficient to load each image one at a time, extract the stars, and then free the image memory.

rmjarvis:
Done as part of #93.

PR #93, opened 2019-12-12
Updates needed for running on DC2

rmjarvis:
A number of updates that I added for running Piff on LSST DC2 images: (cf. https://github.com/LSSTDESC/Piff_DC2)

* Added `satur` values for the input to skip any objects with any pixel that is saturated.  Normally, my input catalog already skips saturated objects, but I was using the truth catalog as the input catalog, so these weren't skipped for me.
* Restructured the input class to read in one file at a time, convert to stars and then throw it out.  This is Issue #82.
* Added `nproc` option to the input field to read and make the stars for multiple input files in parallel.
* Added `nproc` option to the SingleCCD type to do the processing for multiple CCDs in parallel.  (These two values don't need to be the same, since there may be different decision points about memory per cpu to consider.)
* Added an `HSMCatalog` stat type to output a catalog of size and shape values.  The same values that are used for the rho stats (and others), but output into a catalog for further calculations.
* Added input option `hsm_size_reject` to reject stars with very different sizes than the rest.  Since I was using a truth catalog, objects that happened to fall on top of other nearby objects don't work very well, so having this check excluded them from the sample.  The default is a 10 quartile outlier, but that can be adjusted.
* Skip objects that are well off the image sooner, so they don't take a lot of time during input.  This makes it efficient enough to use the full input truth catalog for each image.  It's only a few seconds overhead now removing most of the stars that are not relevant to the image being worked on.


rmjarvis:
Any objection to merging this?  I'll probably also want to tag it 0.4.0 to have a real version number for what I ran on the DC2 images.  If no one speaks up, I'll plan to merge this Monday.

PR #94, opened 2020-04-09
Calculate the params_var for PixelGrid and for GSObjectModel with fastfit

rmjarvis:
We have a statement in the paper about Piff automatically handling the measurement noise properly for gaussian process fitting, so users should not normally need to add the white_noise contribution explicitly.  However, this is currently almost completely aspirational.

The only model that even tries to compute the variance of the params vector is GSObjectModel, and then only when fastfit=False.  However, we don't have any unit tests of this, and it turned out that even that variance estimate was broken.

So this PR adds calculations of the variance for PixelGrid and GSObjectModel(fastfit=True) and fixes the calculation for GSObjectModel(fastfit=False).  For fastfit=True, they are very rough, but hopefully close enough to provide reasonable variance estimate for the GP code.


piff/util.py
```diff
@@ -195,6 +195,28 @@ def hsm(star):
...
+    :param jac:     The Jacobian as a 2d numpy array
+
+    :returns: cov, a numpy array giving the estimated covariance.
+    """
+    import scipy
+    # Cribbed from implemenation in scipy.optimize.curve_fit
+    # https://github.com/scipy/scipy/blob/maintenance/1.3.x/scipy/optimize/minpack.py#L771
+
+    # Do Moore-Penrose inverse discarding zero singular values.
+    _, s, VT = scipy.linalg.svd(jac, full_matrices=False)
```
PFLeget, line 213: Is it safer to do a try/except for the SVD decomposition and return a +infinite in the variance in the same way it is done in pixelgrid?

rmjarvis: I don't think the svd can fail, at least for the same kinds of reasons that the `inv` usage in PixelGrid can fail (e.g. for a singular matrix).  But just in case, I added a try/except block here.  There might be some kind of input matrix that could cause problems, so better safe than sorry.

PR #97, opened 2020-04-20
Add treegp features

PFLeget:
Hi, 

This pull request is the merged of `gp_interp_2pcf.py` & `gp_interp.py` by replacing the used of `sklearn`/`treecorr` by `treegp`/`treecorr` ([https://github.com/PFLeget/treegp](https://github.com/PFLeget/treegp)).

Here are a list of the change: 

- Merge of `gp_interp_2pcf.py`  & `gp_interp.py` into `gp_interp.py` (there is only `GPInterp` now)
- There is an optimizer option that could be `"two-pcf"` or `"log-likelihood"`
- The only dependencies with `sklearn` is by using their kernel implementation 
- All mathematical operation were moved within `treegp` 
- I moved `kernel.py` into `treegp`
- Tests of `GPInterp` are lighter than before because most of the tests are in `treegp` right now
- I removed the `npca` option because I am not sure that this is a GP task, it might be useful, but it should be done by the user outside `GPInterp`. 
- I add an option `rows` in `GPInterp`, like that this is possible to interpolate just the rows from `Star.fit.params` that matters for the GP. It will allow to do only the interpolation on the atmospheric parameters and not on the optical one.  

If you have any question, do not hesitate.
Cheers,
Pierre-François


piff/gp_interp.py
```diff
@@ -13,113 +13,127 @@
...
+                         and anisotropic is True. [default: [3000., 0., 0.]]
+    :param normalize:    Whether to normalize the interpolation parameters to have a mean of 0.
+                         Normally, the parameters being interpolated are not mean 0, so you would
+                         want this to be True, but if your parameters have an a priori mean of 0,
+                         then subtracting off the realized mean would be invalid.  [default: True]
+    :param white_noise:  A float value that indicate the ammount of white noise that you want to
+                         use during the gp interpolation. This is an additional uncorrelated noise
+                         added to the error of the PSF parameters. [default: 0.]
+    :param n_neighbors:  Number of neighbors to used for interpolating the spatial average using
+                         a KNeighbors interpolation. Used only if average_fits is not None. [defaulf: 4]
```
rmjarvis, line 59: This parameter should move below average_fits.  I found it surprising to see here, since I haven't gotten to average_fits yet, so I didn't have context for why you would do a knn interpolation as part of this process.

tests/test_gp_interp.py
```diff
@@ -13,29 +13,37 @@
...
+    for i in range(N*N):
+        if xi[i] !=0:
+            VAR[i] = var[I]
+            I+=1
+        if I == len(var):
+            break
+    VAR = VAR.reshape(N,N) + np.flipud(np.fliplr(VAR.reshape(N,N)))
+    if N%2 == 1:
+        VAR[N/2, N/2] /= 2.
+    return VAR
```
rmjarvis, line 46: I'm pretty confused about this function.  What is the idea here?  Maybe could add some comments saying what the weight is that is output from treegp and what you actually want here as a variance estimate.  The math you're doing here is not at all obvious to me.

Especially please explain the bit where you test `xi[i] != 0`.  When is that expected?  Why would that test be expected to get you the right `I` indexing?

tests/test_gp_interp.py
```diff
@@ -55,1240 +63,258 @@ def get_correlation_length_matrix(correlation_length, g1, g2):
...
+
+    np.testing.assert_allclose(y_test, y_validation, atol = 4e-2)
+
+    if optimize:
+        truth_hyperparameters = np.exp(interp._init_theta)
+        fitted_hyperparameters = np.exp(np.array([gp._optimizer._kernel.theta for gp in interp.gps]))
+        np.testing.assert_allclose(fitted_hyperparameters, truth_hyperparameters, rtol = 2.)
+            
+    if plotting:
+        import pylab as plt
```
rmjarvis, line 182: I don't completely understand why, but apparently using pylab is discouraged.  So probably should switch this to
```
import matplotlib.pyplot as plt
```
which I think is equivalent.

tests/test_gp_interp.py
```diff
@@ -55,1240 +63,258 @@ def get_correlation_length_matrix(correlation_length, g1, g2):
...
+    y_err = np.sqrt(np.array([star.fit.params_var for star in stars_validation]))
+            
+    y_test = np.array([star.fit.params for star in stars_test])
+
+    np.testing.assert_allclose(y_test, y_validation, atol = 4e-2)
+
+    if optimize:
+        truth_hyperparameters = np.exp(interp._init_theta)
+        fitted_hyperparameters = np.exp(np.array([gp._optimizer._kernel.theta for gp in interp.gps]))
+        np.testing.assert_allclose(fitted_hyperparameters, truth_hyperparameters, rtol = 2.)
```
rmjarvis, line 179: rtol=2 is pretty bad.  Is there a reason this doesn't match up very well?  (Should at least give some commentary why we maybe don't expect these to agree to better than a factor of 2 or so.)

.travis.yml
```diff
@@ -34,6 +34,10 @@ before_install:
     - git clone https://github.com/gbernstein/pixmappy.git
     - cd pixmappy; python setup.py install; cd ..
 
+    # Install Treegp
+    - git clone https://github.com/PFLeget/treegp.git
+    - cd treegp; python setup.py install; cd ..
```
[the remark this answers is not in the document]
PFLeget: Yes, I would be happy if you can help me with that task. I never did it before, and I would be happy to learn with some help.

piff/gp_interp.py
```diff
@@ -13,113 +13,127 @@
...
+    :param p0:           Initial guess for correlation length, and quantity of anisotropy
+                         (e1 and e2 params as for galaxy shape). Used only if optimizer is "two-pcf"
+                         and anisotropic is True. [default: [3000., 0., 0.]]
+    :param normalize:    Whether to normalize the interpolation parameters to have a mean of 0.
+                         Normally, the parameters being interpolated are not mean 0, so you would
+                         want this to be True, but if your parameters have an a priori mean of 0,
+                         then subtracting off the realized mean would be invalid.  [default: True]
+    :param white_noise:  A float value that indicate the ammount of white noise that you want to
+                         use during the gp interpolation. This is an additional uncorrelated noise
+                         added to the error of the PSF parameters. [default: 0.]
```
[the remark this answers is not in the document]
PFLeget: I am not sure to understand the comment; you want me to say that the white noise is an uncorrelated additional noise? For the variance I am not so sure, it could be great for the purpose that you proposed, but I am afraid that it could be used in the wrong way.

tests/test_gp_interp.py
PFLeget: I wrote this function for the pull map of the 2d fit. However I don’t need it here, so I removed it.

PFLeget:
@rmjarvis 

I implemented more or less everything you suggested. I improved a lot the testing part, and I added back the`test_yaml()`. The old tests which are not present now compared to what is currently in the master branch of Piff are all the tests of the kernels class which were moved to `treegp` (the test in `treegp` are the same) and the test for the constant PSF and the polynomial PSF. For the constant PSF and the polynomial PSF, I thought it was not a good testing case because if you have a constant/Polynomial PSF pattern,  a polynomial interpolation would be the way of doing it and not a GP. In the previous testing, it was tested only on a single star that the `gp_interp` was tested to be good or not.  Now the testing is done on a validation sample, and I am not sure that the `gp_interp` will be good on the testing sample for a constant/polynomial PSF because they are not described by a GP. But if you want me to add back those test, it will be easy for me to do it. 

If you have any question, do not hesitate.
Cheers,
Pierre-François

.travis.yml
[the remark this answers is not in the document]
rmjarvis: I took the liberty of uploading it to PyPi, so version 0.0.0 is now pip installable.  I had to make a slight change to the setup.py to make it work (it didn't like your classifiers).  If you want, you can add me as a collaborator on that project, and I'll commit the change add some instructions for how to upload future versions.

tests/test_gp_interp.py
```diff
@@ -1077,218 +309,66 @@ def test_yaml():
...
+    np.testing.assert_almost_equal(test_star.fit.params, target.fit.params, decimal=3)
+    # This should also work if the target doesn't have a fit yet.
+    print('interpolate ',piff.Star(target.data,None))
+    test_star = psf.interp.interpolate(piff.Star(target.data,None))
+    np.testing.assert_almost_equal(test_star.fit.params, target.fit.params, decimal=3)
+
+if __name__ == "__main__":
+
+    test_gp_interp_isotropic()
+    test_gp_interp_anisotropic()
```
rmjarvis, line 373: These two tests are taking over a minute each, which is really too long for the regular unit tests runs.
I think they can be sped up using my usual trick of doing something smaller when run by pytest, but the long thing when run from the command line:
```
if __name__ == "__main__":
    nstars = [160, 160, 400, 400]
else:
    nstars = [1600, 1600, 4000, 4000]
```
Sometimes in conjunction with a larger `rtol`.

I'm not sure if `nstars` is the right thing to adjust here, or if so, how small to make it.  So could you think about whether there is an appropriate adjustment you could make on these tests to speed them up for pytest runs?

piff/gp_interp.py
```diff
@@ -13,113 +13,114 @@
...
-            raise RuntimeError("Failed to evaluate kernel string {0!r}.  "
-                               "Original exception: {1}".format(kernel, e))
-        return k
-
-    def _fit(self, X, y, logger=None):
-        """Update the GaussianProcessRegressor with data
+
+        if isinstance(kernel,str):
+            self.kernel_template = [kernel]
+        else:
```
rmjarvis, line 96: This is not covered.  I think it can be removed.  The docs say that kernel should be a str, so as documented, only the True branch is allowed.

piff/gp_interp.py
```diff
@@ -141,6 +142,36 @@ def initialize(self, stars, logger=None):
         :param stars:   A list of Star instances to interpolate between
         :param logger:  A logger object for logging debug info. [default: None]
         """
+        if self.rows is None:
+            self.nparams = len(stars[0].fit.params)
+            self.rows = np.arange(0, self.nparams, 1).astype(int)
+        else:
```
rmjarvis, line 148: This is uncovered.  Probably need a test with rows != None.

piff/gp_interp.py
```diff
@@ -141,6 +142,36 @@ def initialize(self, stars, logger=None):
...
+            self.nparams = len(self.rows)
+
+        if len(self.kernel_template)==1:
+            self.kernels = [self.kernel_template[0] for i in range(self.nparams)]
+        else:
```
rmjarvis, line 153: This is uncovered.  I think it's impossible with only allowing kernel to be a str, given line 95.
Which maybe implies that we don't need it to be a list.

piff/gp_interp.py
```diff
@@ -173,42 +216,84 @@ def interpolateList(self, stars, logger=None):
...
+        self._y_err = np.atleast_1d(data['Y_ERR'][0])
+        self.rows = np.atleast_1d(data['ROWS'][0])
+
+        self._init_theta = init_theta
+        self.nparams = len(init_theta)
+        self.optimizer = data['OPTIMIZER'][0]
+
+        if len(self.kernel_template)==1:
+            self.kernels = [copy.deepcopy(self.kernel_template[0]) for i in range(self.nparams)]
+        else:
```
rmjarvis, line 280: Same here.  I think this branch is impossible now.

piff/gp_interp.py
```diff
@@ -173,42 +216,84 @@ def interpolateList(self, stars, logger=None):
         :returns: a list of new Star instances with interpolated parameters
         """
         Xstar = np.array([self.getProperties(star) for star in stars])
-        y = self._predict(Xstar)
+        gp_y = self._predict(Xstar)
         fitted_stars = []
-        for y0, star in zip(y, stars):
+        for y0, star in zip(gp_y, stars):
             if star.fit is None:
```
rmjarvis, line 222: This branch is not covered.  Is this possible to reach?  Not sure what test to add to hit this line.

.travis.yml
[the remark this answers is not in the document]
PFLeget: Thank you so much for this! I just added you to treegp repo, I except you received an invitation.

piff/gp_interp.py
PFLeget: This is the doc which is not up to date. I updated like that this is possible to have a kernel to be a list of kernel and like that to have a different type of kernel per PSF parameters. I added a test in the `test_gp_interp.py`

piff/gp_interp.py
PFLeget: If kernel is a string at the beginning, it will be only a list with one kernel inside so it will still be used here

piff/gp_interp.py
PFLeget: Same here it can still happened

piff/gp_interp.py
PFLeget: I added a test for that now

PFLeget:
Dear Mike, I followed your recommendation. Let me know if you think that I could improve the coverage.

rmjarvis:
Thanks, Pierre-François!  I rebased to master, so I'll just wait for these final checks to go through (should be identical to the previous version, so should still work), and then merge.

PR #98, opened 2020-05-03
Optatmo minor changes

eacharles:
These are the first set of changes from Ares's development branch.   These are quite localized and very straighforward to review.


piff/des/decaminfo.py
```diff
@@ -144,12 +143,14 @@ def getPosition_chipnum(self, chipnums, ix, iy):
 
         :returns xPos, yPos:    Arrays of x and y coordinates in mm on the focal plane.
         """
-        # do getPosition but with chipnum instead
+        # do getPosition but with chipnum instead
+        if type(chipnums[0]) != int:
```
jmeyers314, line 147: If you just use `chipnums = np.array(chipnums, dtype=int)` once here at the top, can you eliminate the other type conversions below?

piff/config.py
```diff
@@ -49,7 +49,9 @@ def setup_logger(verbose=1, log_file=None):
         handle = logging.StreamHandler()
     else:
         handle = logging.FileHandler(log_file)
-    formatter = logging.Formatter('%(message)s')  # Simple text output
+    formatter = logging.Formatter(
+        fmt='%(asctime)s %(levelname)-8s %(message)s',
+    )
```
rmjarvis, line 54: I'd rather not have this be the default.  Please remove this.  People can always supply their own logger if they want this kind of verbose logging.  

Or as a possible compromise if people really want this, have the time stamped version only when logging to a file, and the simple one when logging to the screen.

piff/des/decaminfo.py
rmjarvis: Yeah, I think this is all simpler with just
```
chipnums = np.array(chipnums, dtype=int, copy=False)
```

piff/config.py
jmeyers314: I think this was originally my commit.  I found the timestamps helpful when working with Ares on finding slow spots in the code.  I wasn't aware that supplying my own logger was an option though; how does that work?

rmjarvis:
```
piff.piffify(config, logger)
```
So, not from the command line version of `piffify`.  But only one short step up from there.

piff/des/decaminfo.py
```diff
@@ -293,8 +294,38 @@ def pixel_to_focal(self, star):
...
         """Take stars and add focal plane position to properties
 
-        :param stars:     Starlist with property 'ccdnum'
+        :param stars:     Starlist with property 'chipnum'
 
         :returns starsl:  New stars with updated properties
         """
         return [Star(self.pixel_to_focal_stardata(star.data), star.fit) for star in stars]
+
+    def get_nominal_wcs(self, chipnum):
```
rmjarvis, line 303: One more thing.  I think this doesn't have any tests currently.  It needs some.

I think I originally wrote this function, so if you can't find relevant tests in the optatmo branch, then let me know and I can write some.

eacharles: I'm not seeing anything.    There is a test_decaminfo test in test_knn_interpy.py, but it doesn't call this function.   Otherwise I'm not anything in tests that even makes DECamInfo objects.

piff/config.py
eacharles: ok, done

piff/des/decaminfo.py
eacharles: ok, and I changed the np.zeros(len(chipnums)) to np.zeros(chipnums.shape)

piff/des/decaminfo.py
rmjarvis: Does anything break if you remove this?  If not, then I guess it's probably leftover cruft from some experimental code, and we don't need it anymore.

eacharles: Well, yes.   It is used in the test scripts in test_optatmo.py in Ares' branch in quite few places.  It doesn't have its own dedicated test, but rather it gets called by the code to make simulated stars for the unit tests. 
We could take it own for now, and add it when we add those tests.   But it seems to me like we could also just figure out now if it is really what we are going to need to test the optatmo_psf code and either put in in or change the test code.

rmjarvis: OK.  I added a quick test that the wcs returned is reasonable.  It only tests one chip that we happen to already have in the input directory, so if we ever find issues where some chips seem off, then we'd probably want to beef this up.  Should at least be good enough now to satisfy codecov though.

Issue #99, opened 2020-05-08
Add weights argument to Star.makeTarget class method.


eacharles:
This would just take that argument and pass it along the the StarData object that is used to construct the Stat.

rmjarvis:
Yep.  Looks like it was just an oversight that this wasn't an option in that function already.

eacharles:
Ok, I'm making at PR for this.

eacharles:
Merged

PR #104, opened 2020-05-14
Added weights argument to star.makeTarget()

eacharles:
This addresses: https://github.com/rmjarvis/Piff/issues/99

### DIFF
--- a/piff/star.py
+++ b/piff/star.py
@@ -159,7 +159,7 @@ class Star(object):
 
     @classmethod
     def makeTarget(cls, x=None, y=None, u=None, v=None, properties={}, wcs=None, scale=None,
-                   stamp_size=48, image=None, pointing=None, flux=1.0, **kwargs):
+                   stamp_size=48, image=None, weight=None, pointing=None, flux=1.0, **kwargs):
         r"""
         Make a target Star object with the requested properties.
 
@@ -186,6 +186,7 @@ class Star(object):
         :param stamp_size:  The size in each direction of the (blank) image. [default: 48]
         :param image:       An existing image to use instead of making a new one, if desired.
                             [default: None; this overrides stamp_size]
+        :param weight:      An existing image to use for the weight. [default: None]
         :param pointing:    The pointing direction to use. [default: None]
         :param flux:        The flux of the target star. [default: 1]
         :param \*\*kwargs:  Additional properties can also be given as keyword arguments if that
@@ -240,10 +241,12 @@ class Star(object):
         image.setCenter(int(x+0.5), int(y+0.5))
         if image.wcs is None:
             image.wcs = wcs
+        if weight is not None:
+            weight.setCenter(int(x+0.5), int(y+0.5))
 
         # Build the StarData instance
         data = StarData(image, image_pos, field_pos=field_pos, properties=properties, 
-                        pointing=pointing)
+                        pointing=pointing, weight=weight)
         fit = StarFit(None, flux=flux, center=(0.,0.))
         return cls(data, fit)
 
--- a/piff/star.py
+++ b/piff/star.py
@@ -242,11 +242,13 @@ class Star(object):
         if image.wcs is None:
             image.wcs = wcs
         if weight is not None:
+            weight = galsim.Image(weight.array.copy(), scale=image.scale)
             weight.setCenter(int(x+0.5), int(y+0.5))
 
+            
         # Build the StarData instance
         data = StarData(image, image_pos, field_pos=field_pos, properties=properties, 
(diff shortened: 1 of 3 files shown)


eacharles:
Hmm.  Do we need to add a test case that includes the weights?

rmjarvis:
> Do we need to add a test case that includes the weights?

Yes.

eacharles:
Hmm.   So the all the new code does is pass the weights argument in star.makeTarget() to the StarData c'tor.  

There isn't really a test that just tests that makeTarget() is doing the right thing.    In test_start.py makeTarget() is only used in the test_io().      

star.makeTarget() is used in a lot of places in higher level tests (test_optics, test_poly, test_wcs).   Is there any particular test or tests where it would make sense to supply weights?

rmjarvis:
It looks like the `make_data` function in test_gsobject_model.py could be reworked to use this feature.  Right now it waits until the `Star` object is made and then fills in the weight image appropriately.  You could switch to passing in a weight to `makeTarget` at the start.

`make_gaussian_data` in test_pixel.py is similar.

eacharles:
Hmm.   I can do something like:
```
    weight = galsim.Image(nside, nside, dtype=float, init_value=1./var)
    star = piff.Star.makeTarget(x=nside/2+nom_u0/du, y=nside/2+nom_v0/du,
                                u=fpu, v=fpv, scale=du, stamp_size=nside, weights=weights)

```
And that works fine for test_gs_object, which isn't really using the weight values
However, looking at similar changes in test_pixel  test_pixel.test_var() I'm concerned
about if the weights image is carrying around the correct metadata.


```
    weight = galsim.Image(nside, nside, dtype=float, init_value=1./var)
    print(weight, weight.array.mean(), var)
    star = piff.Star.makeTarget(x=nside/2+nom_u0/du, y=nside/2+nom_v0/du,
                                u=fpu, v=fpv, scale=du, stamp_size=nside, weight=weight)
    star.image.setOrigin(0,0)
    print(star.data.weight, star.data.weight.array.mean(), var)
    
    g.drawImage(star.image, method='no_pixel', use_true_center=False,
                offset=galsim.PositionD(nom_u0/du,nom_v0/du))
    print(star.data.weight, star.data.weight.array.mean(), var)    
    star.data.weight = star.image.copy()
    star.weight.fill(1./var)
    print(star.data.weight, star.data.weight.array.mean(), var)    

    # To generate output when running the unit test
    assert False # 
```
Produce this output:
```
galsim.Image(bounds=galsim.BoundsI(1,32,1,32), wcs=None, dtype=numpy.float64) 34.60207612456747 0.028900000000000006
galsim.Image(bounds=galsim.BoundsI(0,31,0,31), wcs=None, dtype=numpy.float64) 34.60207612456747 0.028900000000000006
galsim.Image(bounds=galsim.BoundsI(0,31,0,31), wcs=None, dtype=numpy.float64) 34.60207612456747 0.028900000000000006
galsim.Image(bounds=galsim.BoundsI(0,31,0,31), scale=0.5, dtype=numpy.float64) 34.60207612456747 0.028900000000000006

```

eacharles:
I.e., It seems like the makeTarget() function is computing the meta data from the inputs, but adding the weights as it's own object with a possibly conflicting metadata.

eacharles:
is what we want in makeTarget actually somethign like this code below?

```
        if weight is not None:
            weight = galsim.Image(weight.array.copy(), scale=image.scale)
            weight.setCenter(int(x+0.5), int(y+0.5))
```

eacharles:
Ok, I've pushed changes that use the image scale to override the one in the weights object, as above.

rmjarvis:
> weight = galsim.Image(nside, nside, dtype=float, init_value=1./var)

I think this line should include `scale=du`.

eacharles:
Ok, i included that line
However, the star image and weights each carry their own metadata and makeTarget isn't really doing much to ensure that they are consistent.  I'm not sure how much you want to enforce that.

rmjarvis:
Honestly, I don't think it matters. I just looked through the code for uses of weight, and I'm pretty sure we only ever use `weight.array` in anything that matters.  So if the user provides images with inconsistent metadata, it will prefer the image values.  This is a niche enough application, that I don't think it's probably worth adding any checks about this.  I wouldn't be surprised if the only user of this functionality is our own unit tests.

eacharles:
You gonna merge it, or should i?

rmjarvis:
Either way.  :)  Feel free to yourself in the future after an approval.  I usually leave it for the author to merge if they want, in case there is a last minute change they noticed that needs to be fixed or something.

PR #105, opened 2020-05-26
Slight tweaks to GPInterp API

rmjarvis:
I was updating the Piff paper with the latest updates to the GPInterp class (#97), and I think I'd like to propose a slight change in the names for the `optimizer` option.  Specifically `log_likelihood` -> `likelihood` and `two-pcf` -> `isotropic`.

I think this more clearly sets up the default `isotropic` and common alternative `anisotropic` as being very similar, but with the latter allowing for the anisotropy.

And the `log_` part seemed unnecessary as merely an implementation detail about how the maximum likelihood optimization is done.  I think `likelihood` is probably sufficient here.

So here is the text in the paper about the optimizers parameter now:

The TreeGP module has several methods for how to optimize the kernel's
hyper-parameters, which are available in piff via the optimizer parameter:

1. `likelihood` is the traditional maximum likelihood optimization.
    It finds the hyper-parameters that maximize the likelihood of the Gaussian
    process solution.  This is similar to the optimization done by the
    `GaussianProcessRegressor` in scikit-learn although technically 
    TreeGP has a custom implementation, which uses `scipy.optimize` for its back end.
2. `isotropic` uses a direct measurement of the two-point correlation function of the 
    $\hat p_{ik}$ values using TreeCorr. This is the default `optimizer` 
    and usually gives significantly improved results over the maximum likelihood method.
3. `anisotropic` is similar to `isotropic` except that it uses the `TwoD` binning option in 
    TreeCorr, which measures the correlation function is two spatial directions, 
    rather than just radially.  This is particularly useful if the PSF pattern is
    significantly anisotropic (e.g. due to a predominant wind direction).  In this
    case, the resulting kernel is anisotropic as well.
4. `none` does no optimization of the hyper-parameters.  This is probably
    only useful for tests where you may know the true kernel and don't want any
    optimization to be performed.

I also put some of the paper text about GPInterp into the GPInterp doc string as explanation.  And fixed some typos and grammar errors I noticed.

@PFLeget Let me know if this seems ok with you.  I'm close to finalizing the paper, so I'd like to get this merged pretty soon if possible.  Thanks!

### DIFF
--- a/piff/gp_interp.py
+++ b/piff/gp_interp.py
@@ -27,7 +27,28 @@ from .star import Star, StarFit
 
 class GPInterp(Interp):
     """
-    An interpolator that uses Gaussian process from treegp to interpolate multiple surfaces.
+    An interpolator that models the underlying field as a Gaussian process.
+
+    Gaussian process regression, also known as “Kriging,” assumes that the parameters
+    are drawn from a multi-dimensional Gaussian random field across the (u, v) space.
+    It requires an estimate of the spatial covariance function of p, commonly referred to
+    as the kernel.
+
+    The interpolation estimate at an arbitrary location (u, v) is the minimum variance
+    unbiased estimate from the Gaussian distribution at that location conditioned on the
+    values of the parameters measured at all the PSF stars.
+
+    The implemention of this class use the treegp module. (https://github.com/PFLeget/treegp)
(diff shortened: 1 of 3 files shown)


rmjarvis:
Thanks Pierre-François!  I'll update the treegp pip version and make a PR there with information about how you can do so yourself when needed.

PR #108, opened 2020-05-28
Piff 101 input filters

eacharles:
This implements the max_mask_pixels and max_edge_flux additions to the input star filtering.

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -233,9 +233,8 @@ class InputFiles(Input):
                             this value.  [default: 100]
             :max_edge_frac: Cutoff on the fraction of the flux comming from pixels on the edges of the
                             postage stamp. [default: None]
-            :max_mask_pixels: Stars that are partially masked to an extent are cut. Specifically,
-                            stars where at least this many number of pixels have weight 0.0 are
-                            cut [default: None]
+            :max_mask_pixels: If given, reject stars with more than this many masked pixels
+                            (i.e. those with w=0). [default: None]
             :use_partial:   Whether to use stars whose postage stamps are only partially on the
                             full image.  [default: False]
             :hsm_size_reject: Whether to reject stars with a very different hsm-measured size than
@@ -572,11 +571,11 @@ class InputFiles(Input):
         stars = []
 
         if max_edge_frac is not None:
-            edge_mask = np.zeros((stamp_size, stamp_size), bool)
-            for i in range(stamp_size):
-                for j in range(stamp_size):
-                    if np.sqrt(np.square((stamp_size-1.0)/2.0-i)+np.square((stamp_size-1.0)/2.0-j))>(stamp_size-1.0)*(5.0/12.0):
-                        edge_mask[i][j] = True
+            edge_radius = 5./12.                # fraction of stamp_size to use as a radius
+            cen = (stamp_size-1.)/2.            # index at center of array.  May be half-integral.
+            r = edge_radius * (stamp_size-1.)   # radius outside of which is the "edge"
+            i,j = np.ogrid[0:stamp_size,0:stamp_size]
+            edge_mask = (i-cen)**2 + (j-cen)**2 > r**2
         else:
             edge_mask = None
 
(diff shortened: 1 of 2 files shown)


eacharles:
Hi Mike.  Rather than try and deal with all the edge cases for computing the fraction of flux in the edge of the stamps, I just modified the code so that if max_edge_frac is set then it automatically skips stars that only have partial stamps.   I figured it didn't make a lot of sense to have multiple edge cases in dealing with something that was already an edge case.

piff/input.py
```diff
@@ -575,7 +599,7 @@ def _makeStarsFromImage(image_kwargs, cat_kwargs, wcs, chipnum,
                                 "Using smaller than the full stamp size: %s", x, y, bounds)
                 else:
                     logger.warning("Star at position %f,%f overlaps the edge of the image.  "
-                                    "Skipping this star.", x, y)
+                                    "Skipping this star.", x, y)
```
rmjarvis, line 602: Not a big deal, but there's some gratuitous white space added here.  (@jmeyers314 has trained me to avoid this...)

piff/input.py
```diff
@@ -595,6 +619,14 @@ def _makeStarsFromImage(image_kwargs, cat_kwargs, wcs, chipnum,
                 logger.warning("Skipping this star.")
                 continue
 
+            # here we remove stars that have been at least partially covered by a mask
+            # and thus have weight exactly 0 in at least a certain number of pixels of their postage stamp
+            if max_mask_pixels is not None:
+                n_masked = wt_stamp.array.shape[0] * wt_stamp.array.shape[1] - np.count_nonzero(wt_stamp.array)
+                if n_masked >= max_mask_pixels:
+                    logger.warning("Star at position %f,%f has %i masked pixels.",x,y,n_masked)
```
rmjarvis, line 627: Probably ought to say that it is cut because of this.  Otherwise, this just sounds informative.

piff/input.py
```diff
@@ -624,6 +656,18 @@ def _makeStarsFromImage(image_kwargs, cat_kwargs, wcs, chipnum,
...
                 logger.debug("Adding Poisson noise to weight map according to gain=%f",g)
                 star = star.addPoisson(gain=g)
+
+
+            if max_edge_frac is not None and max_edge_frac < 1:
+                flux = np.sum(star.image.array)
+                try:
+                    flux_extra = np.sum(star.image.array[edge_mask])
+                except IndexError:
+                    logger.info("Star at position %f,%f overlaps the edge of the image and max_edge_frac cut is set, skipping this star.")
```
rmjarvis, line 666: Should be logger.warning.

piff/input.py
```diff
@@ -624,6 +656,18 @@ def _makeStarsFromImage(image_kwargs, cat_kwargs, wcs, chipnum,
...
+
+            if max_edge_frac is not None and max_edge_frac < 1:
+                flux = np.sum(star.image.array)
+                try:
+                    flux_extra = np.sum(star.image.array[edge_mask])
+                except IndexError:
+                    logger.info("Star at position %f,%f overlaps the edge of the image and max_edge_frac cut is set, skipping this star.")                    
+                    continue
+                if flux_extra / flux > max_edge_frac:
+                    continue
```
rmjarvis, line 669: Need a logger.warning here also to say skipping this star because x% of flux is near the edge.

eacharles:
Ok, thanks for the suggestions.  I've implemented them all.

piff/input.py
```diff
@@ -556,6 +569,16 @@ def _makeStarsFromImage(image_kwargs, cat_kwargs, wcs, chipnum,
 
         nstars_in_image = 0
         stars = []
+
+        if max_edge_frac is not None:
+            edge_radius = 5./12.                # fraction of stamp_size to use as a radius
```
rmjarvis, line 574: This value still needs to be documented to the user somewhere (i.e. not just an inline code comment).  I think the easiest would be to make it a user-settable parameter, since that provides an obvious place to explain what it means.  But if not that, then either in the explanation of the `max_edge_frac` parameter or in the main section of the doc string, where you could give a long-form (paragraph) explanation of how `max_edge_frac` works.

piff/input.py
rmjarvis: Minor point, but you already calculated flux_frac = flux_extra/flux, so no need to recalculate it here.

Issue #100, opened 2020-05-08
Caching atmospheric params to be more efficienct about making diagnostic plots


eacharles:
In running the diagnostic plots Ares found that code was very slow because it was repeatedly getting the atmospheric parameters for the stars.  

If you look in the eac-dev branch in some of the stats objects you will see that he added flag to signal the optatmo_psf that it did not need to recompute those parameters.   The particular files affected are:

+++ b/piff/star_stats.py
+++ b/piff/stats.py
+++ b/piff/twod_stats.py

And there is code like:
-        self.models = psf.drawStarList(self.stars)
+        if trust_atmo_params_stored_in_stars ==True:
+            self.models = psf.drawStarList(self.stars, trust_atmo_params_stored_in_stars=True)
+        else:
+            self.models = psf.drawStarList(self.stars)

I suspect that Mike has an alternate way of doing this.   
It would be good to agree on how to handle this.   It really speeds up the diagnostic plot making.

rmjarvis:
I don't understand the reason for this.  If the values are already computed, why not just use them.  In other words, what is the use case for `trust_atmo_params_stored_in_stars=False`?

eacharles:
Well, Ares, will need to comment more.   But I think that the root problem is that the star.fit is being overloaded to refer to two versions of the atmospheric fit parameters, (orginal and interpolated) and so when the getParamsList is called, it need to know if it should be using the original version or getting the interpolated version:  the relevant code looks like this:
`                if trust_atmo_params_stored_in_stars:
                    aberrations_atmo_star = np.array([star.fit.params for star in stars])[:,0:3]
                else:
                    # strip star fit
                    stars = [Star(star.data, None) for star in stars]
                    stars = self.atmo_interp.interpolateList(stars)
                    aberrations_atmo_star = np.array([star.fit.params for star in stars])
                params[:, 0:3] += aberrations_atmo_star
`

rmjarvis:
I think the appropriate thing here is to have `drawStarList` check if `star.fit.params is not None` for all the stars being drawn, in which case, don't redo the interpolation.

Doing this for SimplePSF passes all tests, so I think this doesn't break any assumptions we are currently making.  Probably redoing the interpolation was just to make sure things are set.  If they are already set, it's redundant.  Likely whoever originally wrote this (either me or Chris I suspect) didn't anticipate this as an efficiency concern, so we didn't try to optimize it.

Specifically, this diff passes all tests on master:
```
[...]

PR #109, opened 2020-06-03
Don't refit in draw operations if stars already have usable fit parameters

rmjarvis:
Change the behavior of the `psf.drawStar` and `psf.drawStarList` functions in the `SimplePSF` class to not redo the interpolation if the stars already have useable fit parameters.  This is a pretty minor efficiency gain most of the time, since our interpolations are usually pretty quick.  But it is important if the interpolation is GPInterp, since then the interpolation step is relatively expensive.  

In particular, if you compute multiple stats, this only has to do the interpolation once, rather than repeatedly for each statistic, doing the same calculation multiple times.

This is related to an optimization in the optatmo branch discussed on issue #100.  This doesn't directly address that issue -- rather the same adjustment should be made in the optatmo versions of these functions.

### DIFF
--- a/piff/simplepsf.py
+++ b/piff/simplepsf.py
@@ -227,42 +227,84 @@ class SimplePSF(PSF):
 
         logger.warning("PSF fit did not converge.  Max iterations = %d reached.",self.max_iter)
 
+    def interpolateStarList(self, stars):
+        """Update the stars to have the current interpolated fit parameters according to the
+        current PSF model.
+
+        :param stars:       List of Star instances to update.
+
+        :returns:           List of Star instances with their fit parameters updated.
+        """
+        stars = self.interp.interpolateList(stars)
+        for star in stars:
+            self.model.normalize(star)
+        return stars
+
+    def interpolateStar(self, star):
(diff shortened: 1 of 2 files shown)


areshernandez:
Hi Mike,

I agree that at some point in the future making the same adjustment for optatmo is a good idea. However, this will take some time as using the if statement:
``` if star.fit is None or star.fit.params is None:```
will not work since you can have these not being None since the non-interpolated atmo params from the individual star fit are currently saved here, so I will have to come up with some other method.

At any rate, as I've said, this shouldn't be a problem in the near future as I've changed the PIFF fitting pipeline so it only needs to do the interpolation once anyway for each fit.

rmjarvis:
I added `interpolateStar` and `interpolateStarList` methods to force the recalculation.  That's going to be a better interface for forcing the refit than manually deleting params or something (which sounds fragile at best).  See if you like this API.  I think it provides the right hook for any use cases I can think of.

areshernandez:
That makes sense and would be easier to do for optatmo. I'll do something similar there.

PR #110, opened 2020-06-03
Add L0 option to Optical class

rmjarvis:
In the optatmo branch @areshernandez uses a VonKarman profile for the atmospheric component.  This adds this option to the regular `Optical` class simply by specifying L0 as one of the atmospheric parameters.

### DIFF
--- a/piff/optical_model.py
+++ b/piff/optical_model.py
@@ -70,13 +70,15 @@ class Optical(Model):
         :param pupil_plane_im:  The name of a file containing the pupil plane image to use instead
                                 of creating one from obscuration, struts, etc. [default: None]
 
-        Second, there may be an atmospheric component, which uses a galsim.Kolmogorov to
-        model the profile.
+        Second, there may be an atmospheric component, which uses either a galsim.Kolmogorov or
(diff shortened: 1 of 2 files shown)


rmjarvis:
I also changed internal attributes kolmogorov_kwargs -> atm_kwargs since the atm part can now be either kolmogorov or vonkarman.  And optical_psf_kwargs -> opt_kwargs as shorter without (IMO) any loss of clarity.

eacharles:
Hi Mike, I see you assigned this to me, but it looks like you've got it working?   Did you just want me to review it?

rmjarvis:
> Just out of curiosity, why did you change the signal in the test_shearing() test?

Because 1 is a special value here.  That's the starting value of sigma, so if that's the right value, it didn't have to do any work.  1.7 is some random value that isn't special.

In general, I try to make sure any arbitrary values in tests like this aren't 1, even if I don't know that it is special for the particular use case, to try to make sure we don't accidentally hide some bug because 1 does something trivial.  But in this case, I additionally know that 1 is in fact special, so a bad choice for the test.

PR #111, opened 2020-06-05
Add reserve_frac option

rmjarvis:
One last aspirational item that I put into the paper as a "current" feature that we hadn't implemented yet.  This PR adds `reserve_frac` as an optional input field.  This selects a fraction of the input stars as reserve stars and doesn't use them for the psf processing.

These reserve stars can optionally be included in the outlier rejection, but when I tried it out, they were very preferentially selected as outliers.  Maybe not surprising, but it indicates that this might not be a great idea in general.  It's an option, but not the default.  (By default only non-reserve stars are allowed to be rejected as outliers.)

It also implements the reserve nature of the star when outputting the HSMCatalog with shape measurements.  This had already had a placeholder where all stars were marked as not reserved.  This updates that to use the real reserve value.

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -103,14 +103,32 @@ class Input(object):
                       hsm_size_reject=self.hsm_size_reject,
                       max_mask_pixels=self.max_mask_pixels,
                       max_edge_frac=self.max_edge_frac,
-                      stamp_center_size=self.stamp_center_size,
-                      reserve_frac=self.reserve_frac)
+                      stamp_center_size=self.stamp_center_size)
 
-        stars = run_multi(call_makeStarsFromImage, self.nproc, raise_except=True,
-                          args=args, logger=logger, kwargs=kwargs)
+        all_stars = run_multi(call_makeStarsFromImage, self.nproc, raise_except=True,
+                              args=args, logger=logger, kwargs=kwargs)
+
+        # Apply the reserve separately on each ccd, so they each reserve 20% of their stars
+        # (or whatever fraction).  We wouldn't want to accidentally reserve all the stars on
+        # one of the ccds by accident, for instance.
+        if self.reserve_frac != 0:
+            for stars in all_stars:
+                if stars is None or len(stars) == 0:
+                    continue
+                # Mark a fraction of the stars as reserve stars
+                nreserve = int(self.reserve_frac * len(stars))  # round down
+                logger.info("Reserve %s of %s (reserve_frac=%s) input stars",
+                            nreserve, len(stars), self.reserve_frac)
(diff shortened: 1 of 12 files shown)


piff/input.py
```diff
@@ -105,11 +105,30 @@ def makeStars(self, logger=None):
...
+                if stars is None or len(stars) == 0:
+                    continue
+                # Mark a fraction of the stars as reserve stars
+                nreserve = int(self.reserve_frac * len(stars))  # round down
+                logger.info("Reserve %s of %s (reserve_frac=%s) input stars",
+                            nreserve, len(stars), self.reserve_frac)
+                reserve_list = self.rng.choice(len(stars), nreserve, replace=False)
+                # Set them all to False, then update the reserve fraction to True.
+                # This makes sure they all have this value in the properties dict.
+                for star in stars:
```
jmeyers314, line 125: untested, but maybe:
```python
for i, star in enumerate(stars):
    star.data.properties['is_reserve'] = i in reserve_list
```

piff/input.py
```diff
@@ -254,6 +273,13 @@ def __init__(self, config, logger=None):
                             separately to each input catalog.)  [default: None]
             :nproc:         How many multiprocessing processes to use for reading in data from
                             multiple files at once. [default: 1]
+            :reserve_frac:  Reserve a fraction of the stars from the PSF calculations, so they
+                            can serve as fair points for diagnostic testing.  These stars will
+                            have outlier rejection performed along with the regular stars, but
```
jmeyers314, line 278: I'm confused.  I thought you said in the PR description that, by default, the reserve stars will _not_ have outlier rejection applied.

piff/stats.py
```diff
@@ -580,7 +580,10 @@ def compute(self, psf, stars, logger=None):
         self.u = positions[mask, 0]
         self.v = positions[mask, 1]
         self.flux = shapes_truth[mask, 0]
-        self.reserve = np.zeros_like(self.u, dtype=bool)  # For now, all reserve = False
+        self.reserve = np.zeros_like(self.u, dtype=bool)
+        for i,s in enumerate(stars):
+            if s.is_reserve:
```
jmeyers314, line 585: I like `self.reserve[i] = s.is_reserve` here, or maybe even `self.reserve = np.array([s.is_reserve for s in stars], dtype=bool)`

piff/input.py
rmjarvis: Yep.  That's an error.  When I first wrote the code, I thought we'd want that.  But then I made it optional, and by default not.  I failed to update this doc string.

rmjarvis:
Thanks Josh!  I like those better too.

PR #116, opened 2020-06-29
Add center option to PSF.draw method

rmjarvis:
This PR adds a `center` parameter to the `PSF.draw` method, which should be more intuitive for most users in terms of how to use it to specify where the profile should be centered on the resulting image.

As described in #73 the options are:
* `center=None` draws the PSF at the same (x,y) as was used for the interpolation.  This matches the current behavior.
* `center=(xc,yc)` draws the PSF centered at the given (xc,yc) location.
* `center=True` draws the PSF at the exact center of the image.
* `offset=(dx,dy)` still exists and acts as an additional offset to any of the above.

### DIFF
--- a/piff/basis_interp.py
+++ b/piff/basis_interp.py
@@ -385,7 +385,8 @@ class BasisPolynomial(BasisInterp):
         # Start with 1d arrays giving orders in all dimensions
         ord_ranges = [np.arange(order+1,dtype=int) for order in self._orders]
         # Nifty trick to produce n-dim array holding total order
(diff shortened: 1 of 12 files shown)


rmjarvis:
@beckermr This is ready for review if you feel like it.  Something low impact that you can do with a baby on your lap.  :)

Mostly just check that the unit tests cover the kinds of usage that you had wanted Piff to have when you were trying to get things to work a while back.

esheldon:
Assuming the behavior is as the doc string indicates, this looks very good.

rmjarvis:
Great.  Thanks both!

PR #118, opened 2020-07-15
Update des example file

rmjarvis:
This PR updates the example config file in the examples directory to more closely match the one that we used in Y3, which we described in the paper.

It's not completely the same as that file, because I wanted to make it a little easier for people to run a whole exposure, so it uses SingleChip to do all of them in one output file, rather than 60 separate output files.  It also has nicer looking stats images, since it gets to use all the stars in the whole fov.

I also fixed a couple minor issues in the code that turned up when testing this.


rmjarvis:
Results from running this.  First, it takes about 30 min on my laptop.  Much faster on a machine with more cores, since it is mostly parallelized.  (I think close to 60 min of user time, so it effectively did 2x speed on my laptop's two cores.)

Stats plots look pretty reasonable:

D00350139_twod

D00350139_shapes

D00350139_rho

These never look so great on just a single exposure, but nothing terrible standing out here.

jmeyers314:
Looks like we could probably do a better job picking bins in the 1d histograms.  Maybe something like bins=sqrt(n) ?  I'm also curious why there's a difference in the number of x and y bins in the 2d hists.

rmjarvis:
Hm.  Seems that Chris set the defaults to be unequal.  Probably from working on DES chips, which are rectangular.  But the defaults should surely be equal in the two dimensions.  I'll see if I can pretty these up a bit here.

rmjarvis:
Thanks for the prod, Josh.  I worked a bit on prettying up these plots.  And sqrt(n) does seem like a good choice for the number of bins.  I also had to restrict the range to avoid outliers, which were sucking up a lot of nearly empty bins.  Now it just histograms the central 98% of the objects by default, and the example bumps this up to 99.6%, which still manages to avoid the outliers.

I also added the other two stats that we have and tried to improve them as well.  I think they all look nicer now than they did.  (I didn't update Rho at all here, since I had already worked on that earlier this year.)

Here are the current plots now:

D00350139_rho
D00350139_shapes
D00350139_stars
D00350139_twod
D00350139_whisker

Let me know if you have any specific adjustments you think would make them better.

PR #124, opened 2021-08-04
Added color_col option for input Star properties

schutt20:
This adds an input option for a color column and adds 'color' as a valid key to use for interpolation.


rmjarvis:
Hi Theo, I think I'm done futzing with this, so passing it back to you now.  Mostly take a look at the [new unit test](https://github.com/rmjarvis/Piff/pull/124/files#diff-f74f438e9560e140f8f8a9584712f650350696f79733c0def05ee0a4e0edb781R1385).

I did make a couple API changes along the way. 
1. I changed props_cols to property_cols.  A little more verbose, but clearer for the user I think.
2. I got rid of the need for extra_interp_properties at the psf field level.  I realized that this was redundant with the keys in the interp, so we can just get that information from there.

Let me know if this all seems reasonable to you.  If so, we should merge this this week and start the process of making a tagged release.

rmjarvis:
Theo, FYI I rebased this onto the current main to resolve a few conflicts that arose with the other branch that is now merged.  So if you want to make any edits, you should reset to the origin version first:
```
git fetch
git reset --hard origin/add_color_input
```

schutt20:
Hi Mike, thanks for all your work! The unit test makes sense to me, and I appreciate the API streamlining. I don't think I have any additions, so I would be ready to merge.

rmjarvis (approved):

PR #125, opened 2021-08-05
Add `getProfile` to all models.

rmjarvis:
Nominally, this PR is to add a feature where any PSF can get its profile as a GalSim GSObject.  This was pretty easy for most things, since they already used GalSim objects for their internal implementation.  The exception was PixelGrid, which is very similar to the GalSim `InterpolatedImage` class, but it did everything itself, rather than using that class for most of the implementation details.

So the biggest work on this PR was to refactor PixelGrid to use InterpolatedImage natively rather than custom numpy stuff.

The main reason all of this is important is that I think it enables a nice way to implement composite PSF models that we've been talking about.  My idea for them is that the composite class can fit one component at a time and just pass a function `convert_func` that will convert the single component to the full model by adding or convolving (or whatever) the other parts of the complete model before drawing.

So for instance a Sum composite could fit one component at a time using something like
```
for k in range(n_components):
    this_component = components[k]
    others = [c for c in components if c is not this_component]
    convert_func = lambda prof: galsim.Sum([prof] + others)
    component[k].fit(star, logger=logger, convert_func=convert_func)
```
This requires each of the Model fit functions to accept the `convert_func` option and before drawing, just call the function to fold in the other components.  Things would work similarly with Convolve, and I can imagine more complicated composites potentially being useful as well.  You just need to be able to define this function to forward model the process to build up the full profile.

None of the work on the actual composite modeling is part of this PR, but I did stick in the `convert_func` hooks in places to convince myself that it could work.  The hardest one was PixelGrid (of course).  I managed to get something reasonable with a spot for the convert_func to be called before drawing.  But the `chisq` calculation became a fair bit slower this way.  Not that significant really compared to the time taken in the BasisPolynomial solve function, but a factor of ~5 slower than the old chisq code.

So I added a special case there to use a different algorithm (very similar to Gary's old code, but rewritten a bit to make it clearer to me at least) when convert_func is None, and just use the more direct GalSim code when using convert_func.


rmjarvis:
I'd like to get this merged along with Theo's branch and cut a release for running on Y6.  If someone would like to take a look, that would be much appreciated.  But I'll want to merge it next week (let's say Wednesday to be concrete) if I don't hear anything before then.

PR #133, opened 2022-05-16
Fix output T values that had been incorrectly really sigma.

rmjarvis:
@theoschutt noticed a bug in the Piff output catalogs that the columns labeled "T" were really "sigma".  This PR fixes them to be T as advertised according to T = 2 sigma^2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
                 pwd -P
                 cd tests
                 ls -la
-                coverage combine
+                coverage combine || true
                 coverage report
                 ls -la
                 #codecov  # This didn't work.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,3 +12,6 @@ New features
 
 Bug fixes
 ---------
+
+- Fixed bug in output stats that what we called T was really sigma.  Now, it's correctly
+  T = Ixx + Iyy = 2*sigma^2.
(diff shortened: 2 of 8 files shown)


theoschutt (commented):
Thank you for writing this PR. Sorry I never pushed my fix; I was stuck on how to write the tests. I have a comment/question. When I wrote my fix, I converted sigma to T in the Star.run_hsm() method, since that seemed the more fundamental issue. I think the difference only matters if a user would want to use the Star.hsm property directly for some reason. Thoughts?

rmjarvis:
I think it's better to leave star.hsm returning sigma for two reasons.  
1. It's more of an API change.  The hsm function doesn't claim to be returning T, so it's not really a bug.  And it conforms with the GalSim hsm module more closely, so it's likely what users would expect there. 
2. There are other places within Piff where we use the hsm function, and where we do expect it to return sigma, not T.

But I did update the docstring to make it clearer that the "size" it is returning is sigma, not T.

Your question also prompted me to find another place where we treated the hsm output as T - the size-mag selector.  It doesn't matter that much, since it's looking for a locus with constant T, which is also constant sigma.  But to be consistent, I updated it too to make T=2sigma^2.

theoschutt (approved):
Leaving star.hsm to return sigma makes sense. All the other changes look good to me!

rmjarvis:
Thanks Theo!

PR #135, opened 2022-06-15
Fix error in sizemag plot when there are negative flux measurements or no stars

rmjarvis:
@rgruendl found that the size-mag plot could sometimes fail, raising an exception that stops the whole program.  It happened because some hsm fluxes ended up negative, so the magnitude was nan, which caused catastrophic problesm.  This PR fixes this by excluding any objects with negative flux from the plot.  It also now behaves gracefully if there are no stars found or even no input objects (with good hsm measurements) at all.

### DIFF
--- a/piff/size_mag.py
+++ b/piff/size_mag.py
@@ -63,19 +63,19 @@ class SizeMagStats(Stats):
 
         # Pull out the sizes and fluxes
         flag_star = star_shapes[:, 6]
-        mask = flag_star == 0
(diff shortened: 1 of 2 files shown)


theoschutt (approved):
The changes look good to me. Why are you merging into releases/1.2 and not main?

rmjarvis:
> Why are you merging into releases/1.2 and not main?

It's a bug fix, and Robert wants it immediately, so we can release a bugfix release with just this change.

PR #136, opened 2022-06-23
Fix some issues that turned up in Y6 Piff run.

rmjarvis:
The main problem being fixed here is that sometimes for the reserve stars, the hsm fits get influenced by a neighbor and slide off to the side, ending up either with a bad flux or size or both.  Here is an example:
Screen Shot 2022-06-23 at 10 35 28 AM

To fix this, I have the hsm code check for either negative flux or a centroid shift of more than 1 pixel and flag them.  So now the hsm flags are: 1 = HSM itself reported failure, 2 = returned flux < 0, 4 = centroid shift.  These are added bitwise.

While looking into this I fixed a few other things as well:
* When the SizeMag selector calls the initial_select function to get the starting selection of likely stars, it called the normal Select.process function.  This function first selects, but then does some other things including calculate snr values, apply various rejections, and assign reserve stars.  We don't actually want to do any of these other things until we're done with SizeMag.  So now the initial selection stops after just the selection step.
* I noticed that the SizeMag plot that I make was mislabeling the data and models.  Somehow I had these reversed.  So I fixed that.
* I discovered a bug in the StarStat plot when nplot=0, which I fixed.
* I cleaned up the logging a bit where I thought it was being over-verbose on the default verbosity (=1).  I also added some information that I thought was useful to see, but which was absent from the log.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,3 +23,11 @@ Changes from version 1.2.2 to 1.2.3
 
 - Don't skip stars in HSM output file.  Rather just mark failed HSM measurements with
   flag_truth=-1 or flag_model=-1 as appropriate.
+
+Changes from version 1.2.3 to 1.2.4
+-----------------------------------
+
+- Flag objects when HSM returns a negative flux or moves the centroid by more than 1 pixel. (#136)
+- Fixed labeling error in SizeMag plot (model and data were reversed). (#136)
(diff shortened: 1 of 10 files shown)


beckermr (approved):
LGTM to me. I am not a Piff expert so pls take my review with a grain of salt. I didn't see anything that looks out of place.

rgruendl (approved):
Changes look fine to me.  Merge and tag!

PR #138, opened 2022-07-07
Optatmo2.2

aaronroodman:
I have modified optics_model to include a more accurate and flipped/transposed image DECam pupil as well as a map of the Blanco mirror figure.  The unit test in test_optics.py are also modified.

Note that there is a bug in this code - probably a memory leak, since the error is deep in the guts of gals - that occurs when making many stars from this optics model.

The current test_optics.py runs fine, but changing nstars=1000 in the test_makestars method uncovers the problem.


rmjarvis (changes requested):
Just cosmetic changes so far.  I tried running with 1000 stars, and my computer didn't crash, but it did report using 22 GB of memory.  (Macs are pretty good ad swapping memory to disk as needed...)  So I'll try to figure out what's using all that memory and see if I can reduce it.

piff/optical_model.py
```diff
@@ -17,41 +17,69 @@
 .. module:: optical_model
 """
 
+from __future__ import print_function
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -17,41 +17,69 @@
...
 optical_templates = {
-    'des': { 'obscuration': 0.301 / 0.7174,
+    'des_simple': {'obscuration': 0.301 / 0.7174, # from Zemax DECam model
              'nstruts': 4,
-             'diam': 4.274419,  # meters
+             'diam': 4.010,  # meters
              'lam': 700, # nm
-             # aaron plays between 19 mm thick and 50 mm thick
-             'strut_thick': 0.050 * (1462.526 / 4010.) / 2.0, # conversion factor is nebulous?!
+             'strut_thick': 3.5 * 0.019 * (1462.526 / 4010.) / 2.0,  #updated to match big DES donuts, ignores one larger spider leg
```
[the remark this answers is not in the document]
aaronroodman: These are taken from my *donut* code, tuned to match real images of big Donuts.  I kept it as this combination to remember where it came from, but it could certainly just be kept as the final value with the comment that it was tuned to match data.

piff/optical_model.py
```diff
@@ -17,41 +17,69 @@
...
+             'pupil_plane_im': 'input/DECam_pupil_512uv.fits'
            },
 }
 
+gsparams_templates = {
+    'star': {  'minimum_fft_size': 32, 'folding_threshold': 0.02},
+    'starby2': {  'minimum_fft_size': 64, 'folding_threshold': 0.01 },
+    'starby4': {  'minimum_fft_size': 128, 'folding_threshold': 0.005 },
+    'donut': { 'minimum_fft_size': 128, 'folding_threshold': 0.005 },
+    'donutby4': { 'minimum_fft_size': 512, 'folding_threshold': 0.00125 },
```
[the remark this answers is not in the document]
aaronroodman: These are used for optimizing the code for speed.  I found that 'starby2' gave good results for instance.  I believe that to make Donuts - ie. very out of focus - one needs the 'donut' or 'donutby4' setting.

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
                                 it, defined to be positive in the counter-clockwise direction.
                                 [default: 0. * galsim.degrees]
+        :param pad_factor       Numerical factor by which to pad the pupil plane FFT [default: 1]
+        :param oversampling     Numerical factor by which to oversample the FFT [default: 1]
         :param pupil_plane_im:  The name of a file containing the pupil plane image to use instead
                                 of creating one from obscuration, struts, etc. [default: None]
+        :param mirror_figure_im: The name of a file containing an additional phase contribution, for example
+                                 from the figure of the primary mirror. [default: None]
+        :param mirror_figure_halfsize:   The radius of the image containting the mirror_figure_im. [default: None]
+        :param sigma            Gaussian CCD diffusion in arcsec [default:0.]
```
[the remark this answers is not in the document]
aaronroodman: Sounds good

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
+        Second, there is an atmospheric component, which uses either a galsim.Kolmogorov or
         galsim.VonKarman to model the profile.
 
-        :param fwhm:            The full-width half-max of the atmospheric part of the PSF.
-                                [default: None]
-        :param r0:              The Fried parameter in units of meters to use for the Kolmogorov
-                                profile. [default: None]
-        :param L0:              The outer scale in units of meters if desired, in which case
-                                the atmospheric part will be a VonKarman. [default: None]
+        :param atmo_type        The name of the Atmospheric kernel. [default 'VonKarman']
```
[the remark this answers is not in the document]
aaronroodman: which other parameters do you mean?

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+            params[self.idx_z0:self.idx_z0+self.nZ+1] = zernike_coeff[0:selfnZ+1]
+
+        params[self.idx_r0] = r0
+        params[self.idx_L0] = L0
+        params[self.idx_g1] = g1
+        params[self.idx_g2] = g2
+        return params
+
+
+    def getProfile(self,*arguments,**kwargs):
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+        """Get a version of the model as a GalSim GSObject
+
+        :param arguments            An ndarray with params ordered via idx_PARAM data members
```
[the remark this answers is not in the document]
aaronroodman: you mean swap 'arguments' for 'params'?  ok, and yes that comment would be useful

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+        :param kwargs               A dictionary with named parameters: zernike_coeff, r0, L0, g1, g2
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
-
-        :param sigma:           Convolve with gaussian of size sigma. [default: 0]
-        :param g1, g2:          Shear to apply to final image. Simulates vibrational modes.
-                                [default: 0]
+        Finally, there is both an additional Gaussian component to describe CCD diffusion and separately an applied shear.
 
         Since there are a lot of parameters here, we provide the option of setting many of them
         from a template value.  e.g. template = 'des' will use the values stored in the dict
-        piff.optical_model.optical_templates['des'].
+        piff.optatmo_model.aperture_templates['des'].
```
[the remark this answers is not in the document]
aaronroodman: ok, can change name back to optical_templates

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
 
-        :param template:        A key word in the dict piff.optical_model.optical_template to use
-                                for setting values of these parameters.  [default: None]
+        :param template :       A key word in the dict piff.optatmo_model.aperture_template to use
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
+                                for setting values of these aperture parameters.  [default: None]
+        :param gsparams:        A key word in the dict piff.optatmo_model.gsparams_template to use
```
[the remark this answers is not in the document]
aaronroodman: yep

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
+                                for setting values of these Galsim parameters.  [default: None]
 
         If you use a template as well as other specific parameters, the specific parameters will
         override the values from the template.  e.g.  to simulate what DES would be like at
         lambda=1000 nm (the default is 700), you could do:
 
-                >>> model = piff.OpticalModel(template='des', lam=1000)
+                >>> model = piff.OptatmoModel(template='des', lam=1000)
```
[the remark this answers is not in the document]
aaronroodman: yep

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+        return params
+
+
+    def getProfile(self,*arguments,**kwargs):
+        """Get a version of the model as a GalSim GSObject
+
+        :param arguments            An ndarray with params ordered via idx_PARAM data members
+        :param kwargs               A dictionary with named parameters: zernike_coeff, r0, L0, g1, g2
+
+        :returns: a galsim.GSObject instance of the combined OptAtmo PSF
```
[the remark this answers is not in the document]
aaronroodman: The next class to push will be the PSF class, and that is called OptAtmo, but yes, that isn't correct here

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+        """ Get the Atmospheric PSF component.
 
-        if len(prof) == 1:
-            prof = prof[0]
+        : param r0:          Fried parameter
+        : param L0:          Outer scale
+        : param g1:          Shear g1 component
+        : param g2:          Shear g2 component
+
+        : returns: a galsim.GSObject instance
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -176,52 +252,209 @@ def fit(self, star):
...
+                raise ValueError("No value specified for L0 ")
+            try:
+                atm = galsim.VonKarman(lam=self.lam, r0=r0, L0=L0, flux=1.0, gsparams=self.gsparams)
+            except:
+                self.logger.error("Problem making a VonKarman r0,L0 = %f,%f" % (r0,L0))
+        elif self.atmo_type == 'Kolmogorov':
+            atm = galsim.Kolmogorov(lam=self.lam, r0=r0, flux=1.0, gsparams=self.gsparams)
         else:
-            prof = galsim.Convolve(prof)
+            raise ValueError("Invalid atmo_type ",self.atmo_type)
```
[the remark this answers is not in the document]
aaronroodman: ok

piff/optical_model.py
```diff
@@ -67,87 +95,110 @@ def __init__(self, template=None, logger=None, **kwargs):
...
-        :param L0:              The outer scale in units of meters if desired, in which case
-                                the atmospheric part will be a VonKarman. [default: None]
+        :param atmo_type        The name of the Atmospheric kernel. [default 'VonKarman']
 
-        Finall, there is allowed to be a final Gaussian component and an applied shear.
-
-        :param sigma:           Convolve with gaussian of size sigma. [default: 0]
-        :param g1, g2:          Shear to apply to final image. Simulates vibrational modes.
-                                [default: 0]
+        Finally, there is both an additional Gaussian component to describe CCD diffusion and separately an applied shear.
```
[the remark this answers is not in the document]
aaronroodman: correct - the Zernike coefficients & Atmospheric parameters are only taken from the fitted parameters...

aaronroodman:
I am running on Linux - on a Centos7 machine - using Anaconda python 3.8.  I had been using GalSim 2.3.3 and upgraded to 2.3.5, with no change.    I mentioned on SLACK that I pulled out the relevant Galsim calls being used here and just make images with those, and found no memory problem.  That is weird.   So not completely surprised that you are getting different results on a Mac. 
I will implement the above changes and resubmit the pull request, but making 1000 stars and perhaps the Auto build/test will show problems...

piff/optical_model.py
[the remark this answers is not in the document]
aaronroodman: Ah crap - these have probably always been wrong for Galsim - I will fix...

piff/optical_model.py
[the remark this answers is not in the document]
aaronroodman: ok - I will add a sentence to explain this

piff/optical_model.py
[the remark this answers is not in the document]
aaronroodman: I guess that works if I give each one a default value...

piff/optical_model.py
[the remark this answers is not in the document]
rmjarvis: I see.  `donut` and `starby4` are actually the default values.  I guess `star` and `starby2` are basically quick and dirty parameters to run faster at lower accuracy.  You're right that for them, you could potentially need to lower minimum_fft_size in order for the higher folding_threshold to actually end up making a smaller fft image.  So those probably make sense.  But I'd be surprised if the minimum_fft_size in donutby4 is doing anything useful.  I think the lower folding_threshold should be sufficient to get the accuracy you want.

piff/optical_model.py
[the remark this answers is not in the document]
rmjarvis: In the code it still is optical_templates.  If you prefer aperture_templates that's fine too.  But change the name in the code above then.

rmjarvis:
> I will implement the above changes and resubmit the pull request

No need to resubmit.  Just add commits to this branch, and push, and they will be included.  Also, I made a few commits trying to get the GitHub CI to run on a Mac (so far unsuccessfully).  It keeps seg faulting.  I've seen this before in other contexts but not as stubbornly so as this time.  I think they run Macs on a virtual machine or something, and it's a little flaky.  It runs fine on my Mac.

Anyway, you should `git pull` again to grab those commits.  They won't conflict with anything you are doing.

piff/optical_model.py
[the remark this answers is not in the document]
aaronroodman: you are correct about donutby4 - I can remove that one.

rmjarvis:
FYI, I fixed the problem on macos-latest, so those are passing now.  And fixed up the coverage on main, which is now merged in here.  So the coverage failure now is all new code on this branch, which could use some additional unit tests.

piff/optical_model.py
```diff
@@ -61,92 +86,123 @@ class Optical(Model):
...
         if template is not None:
             if template not in optical_templates:
-                raise ValueError("Unknown template specified: %s"%template)
+                raise ValueError("Unknown aperture template specified: %s"%template)
             self.kwargs.update(optical_templates[template])
-        # Do this second, so specified kwargs override anything from the template
+
+        if gsparams is not None:
+            if gsparams not in gsparams_templates:
+                raise ValueError("Unknown gsparams template specified: %s"%gsparams)
```
rmjarvis, line 149: The pattern for testing this is to use `assert_raises` to check that the error is raised when you expect it to be raised.  Either as a function or a context.  Examples:
https://github.com/rmjarvis/Piff/blob/releases/1.2/tests/test_pixel.py#L334
https://github.com/rmjarvis/Piff/blob/releases/1.2/tests/test_pixel.py#L1547

piff/optical_model.py
```diff
@@ -175,52 +256,215 @@ def fit(self, star):
...
+
+        :returns: a galsim.GSObject instance
+        """
+        if self.atmo_type == 'VonKarman':
+            if L0 is None or L0==0.:
+                raise ValueError("No value specified for L0 ")
+            try:
+                atm = galsim.VonKarman(lam=self.lam, r0=r0, L0=L0, flux=1.0, gsparams=self.gsparams)
+            except:
+                self.logger.error("Problem making a VonKarman r0,L0 = %f,%f" % (r0,L0))
```
rmjarvis, line 308: I'm not sure this is right.  Do we really want to catch and ignore (albeit with a logger message) when VonKarmann raises an error?  This is going to lead to another error later when this function tries to return `atm`.  Probably better to just let whatever error galsim raises through so the user can see it rather than swallow it here.
(Also, btw, it's bad form to just use `except` with no types.  When catching errors you should list the errors that you are expecting and know how to handle.  Some like `KeyboardInterrupt` should never be caught.)

piff/optical_model.py
```diff
@@ -175,52 +256,215 @@ def fit(self, star):
...
+            atm = galsim.Kolmogorov(lam=self.lam, r0=r0, flux=1.0, gsparams=self.gsparams)
+        elif self.atmo_type == 'None' or self.atmo_type == None:
+            atm = None
         else:
-            prof = galsim.Convolve(prof)
+            raise ValueError("Invalid atmo_type ",self.atmo_type)
+
+        # shear
+        if g1!=0.0 or g2!=0.0:
+            atm = atm.shear(g1=g1, g2=g2)
```
rmjarvis, line 318: This won't work if `atm=None`.  Maybe easier when not using an atm component to just set it to `DeltaFunction()`.  Convolving with that is equivalent to not convolving with anything.

piff/optical_model.py
```diff
@@ -175,52 +256,215 @@ def fit(self, star):
...
+        :param g2:                 Value of g2 to Shear PSF [default=0.]
+
+        :returns: an ndarray of parameters
+        """
+        params = np.zeros(self.param_len)
+
+        # adapt to a zernike_coeff array shorter or longer than 0:37+1
+        zlen = len(zernike_coeff)
+        if zlen==self.nZ+1:
+            params[self.idx_z0:self.idx_z0+self.nZ+1] = zernike_coeff
```
rmjarvis, line 362: Pretty sure this case would work correctly with either of the below cases, so could merge this into either of them and just have single `zlen<=nZ` or `zlen<=nZ+1` check.  Also, there are apparently no tests currently with zlen > nZ

piff/optical_model.py
```diff
@@ -175,52 +256,215 @@ def fit(self, star):
...
+
+        :returns: a galsim.GSObject instance of the combined optics, atmosphere and diffusion PSF
+        """
+
+        # decode input, if params array or list is present use it alone,
+        # otherwise parameters are in getProfile argument list
+        if len(params)==1 :
+            zernike_coeff,r0,L0,g1,g2 = self.unpack_params(params[0])
+        elif len(params)!=0 :
+            logger.error("getProfile cannot decode arguments:",params)
```
rmjarvis, line 393: I think this should raise an exception (ValueError probably), not just write something to the log.

piff/optical_model.py
```diff
@@ -175,52 +256,215 @@ def fit(self, star):
...
+        # use flux and center properties
+        if use_fit:
+            prof = prof.shift(star.fit.center) * star.fit.flux
+
+        # get image,weight and image_pos from the input star
+        image, weight, image_pos = star.data.getImage()
+        if copy_image:
+            image_model = image.copy()
+        else:
+            image_model = image
```
rmjarvis, line 442: Needs a test.

rmjarvis (changes requested):

piff/optical_model.py
aaronroodman: Added

piff/optical_model.py
aaronroodman: ok - removed this try, except

piff/optical_model.py
aaronroodman: Merged these lines and added a test for zlen>nZ

piff/optical_model.py
aaronroodman: added a test

piff/optical_model.py
```diff
@@ -61,92 +98,123 @@ class Optical(Model):
...
+    :param sigma            Gaussian CCD diffusion in arcsec [default:0.]
 
     Since there are a lot of parameters here, we provide the option of setting many of them
     from a template value.  e.g. template = 'des' will use the values stored in the dict
-    piff.optical_model.optical_templates['des'].
+    piff.optatmo_model.aperture_templates['des'].
 
-    :param template:        A key word in the dict piff.optical_model.optical_template to use
-                            for setting values of these parameters.  [default: None]
+    :param template :       A key word in the dict piff.optatmo_model.aperture_template to use
```
rmjarvis, line 126: This doc is wrong.  It's `piff.optical_model.optical_templates` still.

piff/optical_model.py
```diff
@@ -155,12 +223,37 @@ def __init__(self, template=None, logger=None, **kwargs):
...
+            self.aperture = galsim.Aperture(diam=self.opt_kwargs['diam'],pupil_plane_im=self.opt_kwargs['pupil_plane_im'],
+                                            gsparams=self.gsparams)
+        else:
+            self.aperture = galsim.Aperture(**self.opt_kwargs)
+            #self.aperture = galsim.Aperture(diam=self.opt_kwargs['diam'], obscuration=self.opt_kwargs['obscuration'],
+            #                                nstruts=self.opt_kwargs['nstruts'], strut_thick=self.opt_kwargs['strut_thick'],
+            #                                strut_angle=self.opt_kwargs['strut_angle'],
+            #                                pupil_plane_scale=None, pupil_plane_size=None,
+            #                                oversampling=self.opt_kwargs['oversampling'], pad_factor=self.opt_kwargs['pad_factor'],
+            #                                gsparams=self.gsparams)
```
rmjarvis, line 245: Should delete these commented out lines.

piff/optical_model.py
```diff
@@ -61,92 +98,123 @@ class Optical(Model):
...
+                # build u,v grid points
+                mirror_figure_u = np.linspace(-mirror_figure_halfsize, mirror_figure_halfsize, num=512)
+                mirror_figure_v = np.linspace(-mirror_figure_halfsize, mirror_figure_halfsize, num=512)
+
+                # build the LUT for the mirror figure, and save it
+                mirror_figure_table = galsim.LookupTable2D(mirror_figure_u, mirror_figure_v, mirror_figure_uv.array)
+                self.mirror_figure_screen = galsim.UserScreen(mirror_figure_table)
+
+            # dont put back into opt_kwargs
+            # self.opt_kwargs['mirror_figure_im'] = mirror_figure_im
```
rmjarvis, line 211: Can remove these lines.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
 
         # compute chisq
         chisq = np.std(image.array - model_image.array)
-        dof = np.count_nonzero(weight.array)
+        dof = np.count_nonzero(weight.array) - 6
```
rmjarvis, line 271: Where does 6 come from?  Probably deserves an inline comment.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+    @lru_cache(maxsize=8)
+    def getOptics(self, zernike_coeff):
+        """ Get the strictly Optics PSF component.  Use two phase screens, one from mirror_figure_im, and the other from
+        zernike aberrations
 
-        :param params:      A np array with [z4, z5, z6...z11]
+        : param zernike_coeff:          A ndarray or list with Zernike Coefficients, in units of waves. Array indexed as [0, 0, z2, z3, z4,...z11...]
 
-        :returns: a galsim.GSObject instance
+        : returns: a galsim.GSObject instance
```
rmjarvis, line 285: You have an extra space here.  And on `: param`.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+        """Unpack parameter values
+
+        :param params               An ndarray or list with parameters in order
+
+        :returns: zernike_coeff,r0,L0,g1,g2
+        """
+        kwargs = self.params_tokwargs(params)
+        return kwargs['zernike_coeff'],kwargs['r0'],kwargs['L0'],kwargs['g1'],kwargs['g2']
+
+    def params_tokwargs(self,params):
```
rmjarvis, line 341: More standard name would be `params_to_kwargs`.  Likewise for `kwargs_to_params`.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+            params[self.idx_z0:self.idx_z0+self.nZ+1] = zernike_coeff[0:self.nZ+1]
+
+        params[self.idx_r0] = r0
+        params[self.idx_L0] = L0
+        params[self.idx_g1] = g1
+        params[self.idx_g2] = g2
+        return params
+
+
+    def getProfile(self,*params,zernike_coeff=None,r0=None,L0=None,g1=None,g2=None):
```
rmjarvis, line 383: This is the wrong signature for this function.  This is called by the base class, so it should always just be `getProfile(params)`.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+        :param g1              Atmospheric g1 Shear
+        :param g2              Atmospheric g2 Shear
+
+        :returns: a galsim.GSObject instance of the combined optics, atmosphere and diffusion PSF
+        """
+
+        # decode input, if params array or list is present use it alone,
+        # otherwise parameters are in getProfile argument list
+        if len(params)==1 :
+            zernike_coeff,r0,L0,g1,g2 = self.unpack_params(params[0])
```
rmjarvis, line 399: Rather than use a special function for this, the standard Python way to write this would be
```
*zernike_coeff,r0,L0,g1,g2 = params
```

tests/test_optics.py
```diff
@@ -25,26 +26,28 @@
 @timer
 def test_init():
     # make sure we can init with defaults
-    model = piff.Optical(template='des')
-    prof = model.getProfile([])
+    model = piff.Optical(template='des_simple')
+    prof = model.getProfile(zernike_coeff=[0,0,0,0,0.0],r0=0.15,L0=10.,g1=0.,g2=0.)
```
rmjarvis, line 30: I think this should be `model.getProfile([0,0,0,0,0,0,0.15,10,0,0])` to work with the standard calling syntax.

Or possibly `model.getProfile(model.kwargs_to_params())`.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+        kwargs['g1'] = params[self.idx_g1]
+        kwargs['g2'] = params[self.idx_g2]
+        return kwargs
+
+    def kwargs_toparams(self,zernike_coeff=[0,0,0,0,0],r0=0.15,L0=0.,g1=0.,g2=0.):
+        """Fill params ndarray from kwargs
+
+        :param zernike_coeff:      A ndarray or list with [0, 0, z2, z3, z4, z5, z6...z11..z37] [default=[0,0,0,0,0]]
+        :param r0:                 Value of r0 for Atmospheric Kernel. [default=0.15]
+        :param L0:                 Value of L0 for Atmospheric Kernel. [default=0.0]
```
rmjarvis, line 361: 0.0?  Should this be 10.0 maybe?

tests/test_optics.py
```diff
@@ -254,13 +314,76 @@ def make_empty_star(icen=500, jcen=700, ccdnum=28, params=None,
...
+    for i in range(nstars):
+        # make the shell of a Star object
+        star = make_empty_star(icen[i],jcen[i],chipnum[i],params[i],stamp_size=npixels)
+
+        # draw the star
+        #star = model.draw(star)
+
+        # draw using drawProfile instead of just draw
+        prof = model.getProfile(star.fit.params).shift(star.fit.center) * star.fit.flux
+        star = model.drawProfile(star, prof, params[i], use_fit=True, copy_image=True)
```
rmjarvis, line 366: Why doesn't `model.draw(star)` work?  That's the canonical way to do this.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+
+        # optics
+        prof.append(self.getOptics(tuple(zernike_coeff)))
+
+        # convolve
+        prof = galsim.Convolve(prof,gsparams=self.gsparams)
 
         return prof
 
+    def drawProfile(self, star, prof, params, use_fit=True, copy_image=True):
```
rmjarvis, line 427: This function isn't called by anything except the tests.  Do you really need this special function?  I think the standard functions should work for this.  Either `model.draw(star)` or `psf.drawStar(star)` are the canonical ways to draw onto a star.  Or `image = psf.draw(x,y)` is the thing that most users would use.

piff/optical_model.py
```diff
@@ -175,52 +268,211 @@ def fit(self, star):
...
+        # consider removing that shear by giving this Gaussian opposite shear to the WCS
+        if 'sigma' in self.other_kwargs:
+            if self.other_kwargs['sigma']>0.:
+                gaussian = galsim.Gaussian(sigma=self.other_kwargs['sigma'],gsparams=self.gsparams)
+                prof.append(gaussian)
+
+        # atmospheric kernel
+        atmopsf = self.getAtmosphere(r0, L0, g1, g2)
+        if atmopsf != None:
+            prof.append(atmopsf)
```
rmjarvis, line 417: I don't think `atmopsf` can be None anymore, right?

tests/test_optics.py
```diff
@@ -77,127 +82,169 @@ def test_optical(model=None):
...
+    logger = piff.config.setup_logger(verbose=0)
+    input = piff.InputFiles(config, logger=logger)
+    stars = input.makeStars(logger=logger)
+    astar = stars[10]
+
+    # test drawProfile
+    outparam = [model_kwargs['r0'],model_kwargs['L0'],model_kwargs['g1'],model_kwargs['g2']]
+    aprofile = model.drawProfile(astar,aprof,outparam)
+    aprofile1 = model.drawProfile(astar,aprof,outparam,copy_image=False)
+    aprofile2 = model.drawProfile(astar,aprof,outparam,use_fit=False)
```
rmjarvis, line 132: Should test the standard functions here rather than this one.
But also, what is this test actually testing?  There are no asserts here.

tests/test_optics.py
```diff
@@ -77,127 +82,169 @@ def test_optical(model=None):
...
+    
 
 @timer
-def test_pupil_im(pupil_plane_file='input/DECam_pupil_128.fits'):
+def test_pupil_im(pupil_plane_file='input/DECam_pupil_512uv.fits'):
```
rmjarvis, line 136: Does this need to be a kwarg?  You removed the other calls to this function.  So if there is no reason to call this multiple times with multiple pupil_plane_file's, then this should probably just be a regular variable.

tests/test_optics.py
```diff
@@ -208,13 +255,26 @@ def test_disk():
...
+        assert model.other_kwargs[key] == model2.other_kwargs[key], 'key %r mismatch'%key
+    assert model.atmo_type == model2.atmo_type,'atmo_type mismatch'
+
+@timer
+def test_makestars(nstars=100,constant_atmoparams=True,template='des'):
+
+    print("test_makestars, n=%d, constant_atmoparams=%d" % (nstars,constant_atmoparams))
+    model = piff.Optical(template=template,atmo_type='VonKarman',gsparams='starby2')
+    params = random_params(nstars,model,constant_atmoparams=constant_atmoparams)
+    stars = make_stars(nstars,model,params)
```
rmjarvis, line 276: What is the point of this test?  There are no asserts, and nothing is being done with the final `stars` variable, so I'm not sure what is actually being tested here.

rmjarvis (changes requested):

piff/optical_model.py
aaronroodman: fixed

piff/optical_model.py
aaronroodman: done

piff/optical_model.py
aaronroodman: Thinking about this now, can only identify 3 dof from fitting - for flux and centroid.  Changed to 3 and commented

piff/optical_model.py
aaronroodman: Fixed

piff/optical_model.py
aaronroodman: Thats right. removed the test for None

tests/test_optics.py
aaronroodman: This is to test that the code runs in the mode used by optatmo in a reasonable amount of time, and to compare the timing with different options.   Is there a good way to test for speed, that would work on different machines?

rmjarvis: If that's all it's for, then it shouldn't be in the CI test suite.

Rather, you can put this bit in a `time_optics.py` file, which you could run on different machines to check how long it takes.  Probably living in the devel directory rather than here.  We have a number of that kind of thing in GalSim for instance, which have been very useful for various speed optimizations.

aaronroodman: moved to devel/time_optics.py

tests/test_optics.py
aaronroodman: added test with 2nd pupil file, so leaving as a kwarg

piff/optical_model.py
aaronroodman: This may be historical, but the way I use optical_model in the optatmo PSF class is just via this drawProfile method.  What I do there is the following...

prof = model.getProfile(**model_kwargs) #where model_kwargs have the Zernike coefficients and VonKarman parameters
model_star = model.drawProfile(star,prof,param_array)

neither of the model.draw methods above seem to have a way to get all the needed model parameters for an optical_model, unless there is another way to do this?

aaronroodman: Removed drawProfile

tests/test_optics.py
aaronroodman: removed this from test_optics.py

piff/optical_model.py
rmjarvis: There are still a couple places where aperture_template is used.

rmjarvis:
I think there are only a couple things left to resolve.  I collapsed the comments that I think are done.  

There are also some lines that are not covered by the tests, but I'd be happy to handle that once you're done with everything else, so you can ignore the codecov failure above.

piff/optical_model.py
aaronroodman: this time for sure

piff/optical_model.py
aaronroodman: changed to 10.

piff/optical_model.py
aaronroodman: changed and removed unpack_params method

aaronroodman:
Made changes to address the comments, and if you can add tests for full coverage that would be much appreciated

rmjarvis (approved):

rmjarvis:
Thanks Aaron.  I think we're all done here.  I rebased it onto main.  Once the CI is done to make sure I didn't screw something up with that, I'll merge it.

PR #143, opened 2023-02-04
Add model_properties option to hsm stat output

rmjarvis:
We realized in DES that we need stats about the residuals when the models are not made at the same color as the stars.  This PR adds an option to set any properties you want for the model in stats to be different from the star's own properties.

### DIFF
--- a/piff/stats.py
+++ b/piff/stats.py
@@ -148,7 +148,10 @@ class Stats(object):
 
         :param psf:         A PSF Object
         :param stars:       A list of Star instances.
-        :param model_properties: Optional properties to use for the model rendering.
+        :param model_properties: Optionally a dict of properties to use for the model rendering.
+                                 The default behavior is to use the properties of the star itself
+                                 for any properties that are not overridden by model_properties.
+                                 [default: None]
         :param logger:      A logger object for logging debug info. [default: None]
 
         :returns:           positions of stars, shapes of stars, and shapes of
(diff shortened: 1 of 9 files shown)


piff/stats.py
```diff
@@ -689,9 +689,8 @@ def compute(self, psf, stars, logger=None):
         # Also write any other properties saved in the stars.
         self.props = {}
         prop_keys = list(stars[0].data.properties)
-        # Already did the position ones.
-        for key in [ 'x', 'y', 'u', 'v' ]:
-            prop_keys.remove(key)
+        # Remove all the position ones, which are handled separately.
+        prop_keys = [key for key in prop_keys if key not in ['x', 'y', 'u', 'v', 'ra', 'dec']]
```
theoschutt, line 698: Just curious: why is this way better (besides including 'ra' and 'dec')?

tests/test_star.py
```diff
@@ -371,6 +371,10 @@ def test_star():
     assert star.flux == 1.
     assert star.center == (0,0)
     assert star.is_reserve == False
+    assert star['color_ri'] == 0.5
+    assert star['color_iz'] == -0.2
+    with np.testing.assert_raises(KeyError):
```
theoschutt, line 376: When designing a unit test, how do you decide which exceptions to test for? It seems like there could be also be a `TypeError` when making a Star object, for example.

theoschutt (approved):
Looks good to me!

piff/stats.py
rmjarvis: Because ra, dec aren't necessarily in there.  And python doesn't have a very easy "remove if present" for lists like it does for dicts and sets (`d.pop(key, None)`)

tests/test_star.py
rmjarvis: At a minimum I try to include any errors that we explicitly raise in our code.  That doesn't apply in this case though.  This one was just because later I'll add color_gi using 'withProperties', so checking that it's not present here was just to add some parallelism to the two cases.

Issue #142, opened 2023-01-20
Add optional high order moments to the HSM output file.


rmjarvis:
@ztq1996 has made a convincing case that we should look at the high-order moments of the PSF to look for potential biases in the PSF models leading to biased galaxy shapes.

This would be especially easy for many use cases if we add these moment measurements to the standard hsm output catalog we make in Piff.  My proposal is to add a `fourth_order=True` option to the HSMCatalog statistic type, which would turn on these measurements.  

To make these more directly useful to galaxy shape error diagnostics, I propose that we output these in terms of the spin-0, spin-2 and spin-4 combinations:

spin-0: M_40 + 2M_22 + M_04.
spin-2: (M_40 − M_04) + i(2M_31 + 2M_13)
spin-4: (M_40 - 6M_22 + M_04) + i(4M_31 - 4M_13)

The spin-4 one probably isn't too useful (e.g. there is no spin-4 quantity in TreeCorr to do proper correlations with it), but the other two definitely are.

TQ calls the first two rho and M^(4)_PSF, respectively, which I don't love as names for these.  They are kind of the 4-th order T and e quantities, so maybe T4 and e4 would be more accessible to most users?  I'm open to suggestions here.

We also need some name for the spin-4 quantity (TQ didn't name it AFAICT).

There is already a function in Piff to calculate the high-order moments, so I think this will be easy to add.  Just call that function if fourth_order is set to true and add those columns to the output file.  So the implementation should be pretty straightforward.

Refs: 
https://ui.adsabs.harvard.edu/abs/2022MNRAS.510.1978Z/abstract
https://ui.adsabs.harvard.edu/abs/2022arXiv221203257Z/abstract

ztq1996:
Hi @rmjarvis - 

Thanks for pointing this out. I feel like we have touched on this a few times verbally but never written something down. 

I just took a look at `piff.util.calculate_moments()`. It looks pretty good to me for setting up higher moments measurement. I just  have a few questions and comments:
1. I have a hard time lining up the index of M_{pq} in piff with its mathematical meaning: for example, for second moments, M11=trace, M20=e1, M02=e2. I don't think we need to change how the indices are done here, as that could break some other code, but maybe we need some explanation for the indices.
[...]

PR #144, opened 2023-02-24
#142 Add optional fourth_order=True and raw_moments=True options for HSMCatalog

rmjarvis:
This PR adds two options to include higher-order moments in the HSM output file.

The first one, fourth_order=True, is the one I expect to be more useful in practice.  It outputs additional columns: T4, g41, g42, h41, h42. 

T4 is a spin-0 quantity with units of area (like T) based on fourth order moments.  Specifically T4 = <r^4> / T.  For a round Gaussian, it comes out equal to 2T.  For an elliptical Gaussian, there is a correction of approximately (1-e^2)^0.5.

g4 is a spin-2 dimensionless quantity (like g).  For this one, we subtract off the leading effect of the overall ellipticity (=3e), so g4 is really probing some additional spin-2 effect beyond the trivial geometry already indicated by the regular second-moment shape, g.  For an elliptical Gaussian, this is close to 0.  (Thanks for @ztq1996 for this suggestion.)

h5 is a spin-4 dimensionless quantity.  I'm not sure what the uses for this might be, but it seemed worth including.

The second optional arguments is raw_moments=True.  This is simply the raw output of calculate_moments, outputting all the moments that it can calculate without any scalings or correction terms.  @aaronroodman thought this would be useful for checking the accuracy of the optical PSF fits, to check things like trefoil and such are accurately modeled.

Along the way, I discovered that the calculation of T from sigma that we did in #133 wasn't quite right.  The sigma that HSM outputs is the fourth root of the determinant of the moment matrix, not the square root of half the trace.  So there is a correction required involving the measured shape to get the definition of T that we want (i.e. the trace of the moment matrix).  So I fixed that here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,9 @@ jobs:
 
         steps:
             - uses: actions/checkout@v2
+              with:
+                # Helpful for a reliable codcov upload.
+                fetch-depth: 0
 
             - name: Set up Python ${{ matrix.py }}
               uses: actions/setup-python@v2
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,4 +27,4 @@ Bug fixes
 ---------
 
 - Fixed bug in output stats that what we called T was really sigma.  Now, it's correctly
-  T = Ixx + Iyy = 2*sigma^2. (#133)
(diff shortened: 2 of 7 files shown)


theoschutt (approved):
See in-line comments. They're just clarifications/typos, so I'll approve now.

Issue #141, opened 2022-12-14
coadd_object_id in psf_stars extension is stored as an f8 instead of an i8


esheldon:
COADD_OBJECT_ID    >f8  1698368275.0

esheldon:
chipnum is also float rather than int

esheldon:
These also should be int GAIA_SOURCE_ID, VHS_SOURCEID, FLAG_COLOR, is_reserve

rmjarvis:
We currently save all the extra properties we take from the input catalog as float in the output.  The fix would be to keep track of what type they were in the input catalog and copy that to the output.

Issue #140, opened 2022-12-14
incorrect error message with missing keyword


esheldon:
If you don't send chipnum it says that you have sent an unexpected keyword, but then printing the message fails

This should I think say there is a *missing* keyword, namely chipnum

```python
Cell In[24], line 1
----> 1 im = p.draw(35, 40, center=True, GI_COLOR=1)

File ~/miniconda3/lib/python3.9/site-packages/piff/psf.py:170, in PSF.draw(self, x, y, chipnum, flux, center, offset, stamp_size, image, logger, **kwargs)
    103 r"""Draws an image of the PSF at a given location.
    104
    105 The normal usage would be to specify (chipnum, x, y), in which case Piff will use the
   (...)
    166 :returns:           A GalSim Image of the PSF
    167 """
    168 logger = galsim.config.LoggerWrapper(logger)
--> 170 prof, method = self.get_profile(x,y,chipnum=chipnum, flux=flux, logger=logger, **kwargs)
    172 logger.debug("Drawing star at (%s,%s) on chip %s", x, y, chipnum)
    174 # Make the image if necessary

File ~/miniconda3/lib/python3.9/site-packages/piff/psf.py:252, in PSF.get_profile(self, x, y, chipnum, flux, logger, **kwargs)
    249     raise TypeError("Unexpected keyword argument(s) %r"%list(kwargs.keys())[0])
    251 image_pos = galsim.PositionD(x,y)
--> 252 field_pos = StarData.calculateFieldPos(image_pos, self.wcs[chipnum], self.pointing,
    253                                        properties)
    254 u,v = field_pos.x, field_pos.y
    256 wcs = self.wcs[chipnum]

KeyError: 0


```

PR #145, opened 2023-02-27
Three quick fixes based on DES Y6 usage

rmjarvis:
This PR fixes three things we ran into when using Y6 Piff files:

1. Issue #141: Erin noticed that the HSM output files make most things float, including things read from the input file, which should really be ints.  Now we keep track of the dtype of any properties that are read from the input file and use the same dtype in the output.
2. Issue #140: If a PSF model only covers a single CCD, then the requirement about whether to specify that chipnum when calling `psf.draw` depends on how the piff config file was written.  I.e. whether a chipnum was specified there or not.  The final model is the same either way, but you get an error if you do it the wrong way.  Now if the model is just for one chip, you don't need to specify the chipnum regardless of how the config file handles the chipnum.
3. We discovered that the default Lanczos(3) interpolant for PixelGrid is not particularly accurate if you always draw in the center of a pixel, rather than either randomly or at the original star location.  Bumping this up to Lanczos(7) made the inaccuracies negligible, so we are switching the default interpolant to that.


piff/input.py
```diff
@@ -979,8 +984,6 @@ def safe_to_image(wcs, ra, dec):
                     raise KeyError("Key %s not found in FITS header"%gain)
                 gain = float(header[gain])
             extra_props['gain'] = np.array([gain]*len(cat), dtype=float)
-        else:
-            extra_props['gain'] = [None] * len(cat)
```
beckermr, line 983: I don't understand this block, not because I think it is wrong, but because I don't know this code base. I am pointing this out to say that I am not qualified to review this bit I think.

piff/input.py
```diff
@@ -996,8 +999,6 @@ def safe_to_image(wcs, ra, dec):
                 satur = float(header[satur])
                 logger.debug("Using saturation from header: %s",satur)
             extra_props['satur'] = np.array([satur]*len(cat), dtype=float)
-        else:
-            extra_props['satur'] = np.array([np.inf]*len(cat), dtype=float)
```
beckermr, line 1000: I don't understand this block, not because I think it is wrong, but because I don't know this code base. I am pointing this out to say that I am not qualified to review this bit I think.

beckermr (commented):
LGTM to the best of my knowledge! 

There were two lines whose removal I did not follow. I also am definitely a bit fuzzy on the changes to the gain and satur stuff in general. I think that section could use a second look if there is is someone since I don't know those code bits at all.

piff/input.py
rmjarvis: We used to always put gain in extra_props, even if it wasn't set by any of the various input options.  Now that we are writing all the properties to the HSM catalog, these show up as NaN, which is ugly and potentially confusing to users.  Same with satur.
So I changed it to not include them in the extra_props dict if they weren't defined, which just slightly changed how I check for that in the code.  But that means they don't end up getting written to the HSM catalog.

beckermr: Thanks for the explanation. Code looks good for that!

rmjarvis:
Thanks Matt!

PR #146, opened 2023-03-06
Add Tutorial notebook

rmjarvis:
Last Friday, I did a DESC onboarding sprint giving a tutorial of how to use Piff.  Both building models and using the models.  The tutorial was in the form of a Jupyter notebook, so I'm adding it to the repo here.

I've also added automatic testing of the notebook in our CI to make sure that the notebook stays correct with any future updates to Piff.


examples/Tutorial.ipynb
```diff
@@ -0,0 +1,899 @@
...
+   "metadata": {},
+   "source": [
+    "### Overview\n",
+    "\n",
+    "Piff is primarily designed to be run from the command line using a configuration file.  The (primate) executable is called `piffify`, so the most straightforwared usage would be to run e.g.\n",
+    "```\n",
+    "piffify config.yaml\n",
+    "```\n",
+    "\n",
+    "It is also typical to change individual configuaration items on the command line that are meant to change for each exposure or each input image.  So perhaps a more typical usage would look something like\n",
```
plazas, line 32: Typo: "configuaration"

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "84215581",
+   "metadata": {},
+   "source": [
+    "### Overview\n",
+    "\n",
+    "Piff is primarily designed to be run from the command line using a configuration file.  The (primary) executable is called `piffify`, so the most straightforwared usage would be to run e.g.\n",
```
plazas, line 27: Typo: "straightforwared"

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "\n",
+    "config = \"\"\"\n",
+    "\n",
+    "input:\n",
+    "    \n",
+    "    # dir is optional.  If set, it applies to all file names here without an absolute path\n",
+    "    dir: 572501\n",
+    "        \n",
+    "    # Define the input image with the pixel data.\n",
+    "    # These data come with a bad pixel mask and a weight map.\n",
```
plazas, line 100: Is the weight map expected to be an inverse variance map? In which units? Should the variance also include the shot noise from the source? 
And for the bad-pixel map, is there any particular expected form? (e.g., "1" for good pixels, "0" for bad; or different mask planes like in the LSST Science Pipelines?)

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "    # The default is to use all input objects as candidate stars.\n",
+    "    # Here we refine that a bit by selecting rows based on the SExtractor \"SPREAD_MODEL\" value.\n",
+    "    type: Properties\n",
+    "        \n",
+    "    # This can be any eval-able string.\n",
+    "    # Here we use the SPREAD_MODEL property we read in from the input catalog.\n",
+    "    where: \"np.abs(SPREAD_MODEL) < 3.e-4\"\n",
+    "        \n",
+    "    # Mark 20% of stars as \"reserve\" stars.  Don't use them for fitting.                     \n",
+    "    # In ML lingo, this would be a validation set.\n",
```
plazas, line 134: Since this is a pedagogical tutorial, it'd be nice to define all acronyms like "ML", "Sextractor", and "DES" the first time they're used.

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9558ca81",
+   "metadata": {},
+   "source": [
+    "For some use cases, it may be preferable to access the PSF as a GalSim GSObject.  E.g. for forward modeling, you could then convolve it directly with your galaxy model.  Or you may want to draw it directly in Fourier space rather than in real space and then apply an FFT.\n",
+    "\n",
+    "The `get_profile` function doesn't just return the profile.  It returns a tuple `(profile, method)`.  The `method` is either 'fft' or 'no_pixel', corresponding to the method one would want to use if drawing this profile onto an image using the galsim `drawImage` method.  This applies either to just directly drawing the PSF or first convolving the PSF by a galaxy (or other object profile) and then drawing.  \n",
```
plazas, line 421: ``galsim``-->``GalSim``

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7d1dcb07",
+   "metadata": {},
+   "source": [
+    "# Piff Tutorial\n",
+    "\n",
+    "### Links\n",
```
plazas, line 10: Maybe at the top, you could add a section/header like the ones in the DP0.2 tutorials, which include information such as the version of the code used to run the notebook ad the last time it was verified to run: https://github.com/rubin-dp0/tutorial-notebooks/blob/main/01_Introduction_to_DP02.ipynb

examples/Tutorial.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "\n",
```
plazas, line 11: As a general comment, I recently put together some recommendations to make Jupyter Notebooks more accessible. Can you access the following page?: https://confluence.lsstcorp.org/pages/viewpage.action?pageId=192909126 
If so, at the bottom there is a section called "Jupyter Notebooks resources for vision-impaired astronomers". I know it seems like a lot and I'd be happy to discuss it further with you if you want :)  There are recommendations such as: " Use colorblind intuitive palettes for graphics. As pointed out above, we may already be doing this in some tutorials (e.g., NB03), but we should check them all. 
Maximize accessibility for people with dyslexia. Use dyslexic-accessible fonts (including but not limited to Helvetica, Verdana, and OpenDyslexic) at a reasonably large font size (12 pt = 16 px) with sufficient space between lines. Avoid crowding, and follow best practices related to the length and arrangement of 2 sentences that are displayed. Many potential astronomy students have dyslexia 3 (Schneps 2006), and some people with disabilities are excluded because they feel incompetent when the situation is mostly related to access to information."

examples/Tutorial.ipynb
rmjarvis: Yep.  LSST does it different from the usual kind of weight map.  They report variance, not inverse variance, and also include the signal as part of that variance, which we don't want.  So there are options to convert it.
```
invert_weight: True
remove_signal_from_weight: True
```

rmjarvis: I think I mentioned these verbally, but I'll add something about that in the tutorial.

examples/Tutorial.ipynb
rmjarvis: It's being run as part of the CI tests, and it's hosted in the PIff repo itself, so it will be kept current with the current version of Piff, and there will be a version connected to each minor release series, which would be known to work with that version.

examples/Tutorial.ipynb
rmjarvis: I think everything is already pretty high contrast, and I think the font is already Helvetica or something similar.  (Also I don't actually know how to change it.)  Are there any specific changes you'd recommend related to this?

rmjarvis:
@theoschutt e42c9562025

PR #149, opened 2023-04-10
Guard against bad HSM fit when fourth_order=True

rmjarvis:
We forgot to guard against the possibility of bad hsm fits (which are relatively common actually) when doing the fourth order moments in the HSM output file.  This fixes it.


theoschutt (commented):
Looks good to me. For the unit test, do we need to check more about the output stats file, e.g. that higher order stats got set to 0?

rmjarvis:
I think if it's flagged, we don't care what the high order stats were set to.  The row just needs to be the right length, which is why we extend with the right number of 0's.

PR #151, opened 2023-04-15
Change how we register the various valid type names for the config

rmjarvis:
I've never been super happy with how we check for valid type names in the config dict, since the names were required to be connected to the class names, but not in a particularly consistent way.

So before I start adding even more type names for the composite PSFs, I wanted to update how this is done.  Now there is a dict for each base class (PSF, Model, Interp, Outliers, Input, Output, Select and Stats) called e.g. valid_psf_types, etc., which keeps track of the config names and the corresponding class names to use.

Each class declares their config type name with a class variable called `_type_name`, which gets registered in the dict automatically using the `__init_subclass__` method.  That's a python 3-only feature, so this implementation wasn't possible when we were still supporting python 2.7.  (At least not as easily.)

I also changed a couple of rarely used type names, which I thought were a bit awkward or not very clear.  The old names still work, but are now deprecated (until the eventual Piff version 2.0):
  * GSObjectModel -> GSObject
  * GPInterp -> GP or GaussianProcess
  * kNNInterp -> KNN or KNearestNeighbors
  * Star -> StarImages  (a Stat type)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ API Changes
   * GSObjectModel -> GSObject
   * GPInterp -> GP or GaussianProcess
   * kNNInterp -> KNN or KNearestNeighbors
+  * Star -> StarImages
 
 
 Performance improvements
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,8 +9,10 @@ Output file changes
 API Changes
 -----------
 
-- Deprecated some redundant type names.  GSObjectModel -> GSObject,
-
+- Deprecated some redundant or potentially unclear type names.
+  * GSObjectModel -> GSObject
+  * GPInterp -> GP or GaussianProcess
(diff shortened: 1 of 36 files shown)


theoschutt (approved):
```
# This class-level dict will store all the valid model types.
# Each subclass should set a cls._type_name, which is the name that should
# appear in a config dict.  These will be the keys of valid_model_types.
# The values in this dict will be the Model sub-classes.
```

Might want to include this explanation in every base class. Looks like `Outliers`, ` Input`, `Output`, `Stats`, and `Select` don't have it yet. Currently, I could see someone writing, e.g., a new Stats subclass and never coming across the directions to set a `cls._type_name` (until they get the helpful exception message).
Otherwise looks good to me!

rmjarvis:
Good idea.  Thanks Theo.  I started that way, but then got lazy.  :)  Good to call me on that.  They all have the comment now.

PR #152, opened 2023-04-20
optatmo_psf wavefront based PSF

aaronroodman:
This pull request has optatmo_psf included.  Note that the only unit test included just makes the PSF class, nothing more is implemented yet.

In the devel area are two scripts with high level but lengthy tests of optatmo_psf:

`python time_optatmo_mkimage.py psf_optatmo_time.yaml` uses the PSF class to make a focal plane's worth of stars, writing these out to a pickle file.  By default no extra atmosphere is added, only a constant atmosphere.

`python time_optatmo_fit.py psf_optatmo_time.yaml` then reads in the stars and fits them.  The fit takes 45 minutes on an old linux machine, and should yield fit results with the same values as used to make the stars.  There is no checking implemented, only printout.

I don't see where to label this PR a work in progress, but obviously thats what this is.

I think it would be great if you can give optatmo_psf a high level review, especially to see if it is implementing the PSF interface correctly.  Issues that I know need work are a) how to store the optical fit parameters and b) how to write out the stars into a PIFF file and c) how to write out the fit output in a PIFF file.

### DIFF
--- a/devel/piff_test_helper.py
+++ b/devel/piff_test_helper.py
@@ -1,0 +1,1 @@
+../tests/piff_test_helper.py
--- a/devel/psf_optatmo_time.yaml
+++ b/devel/psf_optatmo_time.yaml
@@ -1,0 +1,66 @@
+# Config file for optatmo_psf testing
+
+modules:
+   - galsim_extra
+
+output:
+  file_name: psf_optatmo.piff
+
+psf:
+
+  type: OptAtmo
+
+  model:
+     type: Optical
+     lam: 782.1
+     template: des_128
+     gsparams: starby2
+     atmo_type: 'VonKarman'
+
+  interp:
+     type: GPInterp
+     #anisotropic: true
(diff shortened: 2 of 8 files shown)


aaronroodman:
Mike,  

you had made some edits to devel/time_optatmo_mkimage.py, which included removing this code:

```
        # get 'fit' parameters for this star
        params = list(opt_params.copy())
        if atmo_type!='Const':
            params.append(init_params.get('atmo_size_%d' % (i)))
            params.append(init_params.get('atmo_g1_%d' % (i)))
            params.append(init_params.get('atmo_g2_%d' % (i)))

        fit = StarFit(params,
                      flux=star.fit.flux,
                      center=star.fit.center)
        noisy_star = Star(data, fit)
```

params has all the information about how this star was made, and my understanding is that belonged in the StarFit object as above.  Is that not correct, and if not where should that information go

PR #155, opened 2023-05-30
Add init and fit_flux options to models

rmjarvis:
This is the last PR in preparation for the introduction of composite PSF models.  The changes here include:

1. Made it possible for `star.fit.params` to be a list of np.arrays for multiple components.  These are accessible using `get_params(num)` when there are multiple models being parametrized.  This PR doesn't introduce any such models yet, but they are coming...  :)
2. Added an `init` option to models.  Currently all models start with something close to an hsm fit to the star as an initial guess.  That's not always a good choice for composite models, since you don't want every component to be similar to the whole star.  Now there are 3 values of init that every model should accept: 
   * `hsm`, which matches the current behavior and is the default
   * `zero`, which means start as something close to zero -- useful for Sum components designed to mop up systematic residuals of some kind
   * `delta`, which means start as something close to a delta function -- useful for Convolve components designed to be a small convolutional correction.
3. Added another `init` option for GSObjectModel: a tuple `(flux, size)`, which represent flux and size scalings relative to the hsm fit.  I found this useful when testing Sum types when multiple components are all significant fractions of the total size or flux to speed up convergence of the fit.
4. Added `fit_flux` options to GSObject and PixelGrid models.  Currently models are always assumed to have flux=1, and the star.fit.flux then captures the flux scaling for any given star.  However, Sum PSFs will need some components to have flux != 1, so the relative fluxes can vary between components (and the total flux would still be normalized to 1).  For GSObject and PixelGrid, this is fairly straightforward, so they are allowed to take on this kind of role.
5. Deprecated the `copy_image=False` option for drawStar and drawStarList.  This option violates our usual immutable star paradigm, and it was never used except in our tests, so I think it's not worth keeping.

### DIFF
--- a/piff/gsobject_model.py
+++ b/piff/gsobject_model.py
@@ -331,8 +331,13 @@ class GSObjectModel(Model):
         if convert_func is not None:
             prof = convert_func(prof)
 
-        prof.drawImage(model_image, method=self._method, center=star.image_pos)
-        chisq = np.sum(star.weight.array * (star.image.array - model_image.array)**2)
+        try:
+            prof.drawImage(model_image, method=self._method, center=star.image_pos)
+        except galsim.GalSimError:
+            # If we get the exception here, then set a large chisq, so it gets outlier rejected.
+            chisq = 1.e300
+        else:
+            chisq = np.sum(star.weight.array * (star.image.array - model_image.array)**2)
         # Don't subtract number of parameters from dof, since we'll be interpolating, so
         # these parameters don't really apply to each star separately.
         # After refluxing, we may drop this by 1 or 3 if adjusting flux and/or centroid.
--- a/piff/gsobject_model.py
+++ b/piff/gsobject_model.py
@@ -65,6 +65,7 @@ class GSObjectModel(Model):
             self._nparams = 3
         else:
             self._nparams = 5
+        self.set_num(None)
 
     def moment_fit(self, star, logger=None):
         """Estimate transformations needed to bring self.gsobj towards given star."""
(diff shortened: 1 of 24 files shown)


piff/gsobject_model.py
```diff
@@ -289,27 +337,60 @@ def fit(self, star, fastfit=None, logger=None, convert_func=None):
...
+        if not self._raw_size > 0:
+            raise RuntimeError("Error computing the size of the reference object")
+        size /= self._raw_size
+
+        flux_scaling = None
+        if init == 'zero':
+            flux_scaling = 1.e-6
+        elif init == 'delta':
+            size *= 1.e-6
+        elif init.startswith('(') and init.endswith(')'):
```
jmeyers314, line 369: From the docstring, I expected this option to be a literal tuple rather than a stringified tuple.  Is there a reason this should be a string?

piff/star.py
```diff
@@ -941,22 +961,72 @@ def alpha(self):
     def beta(self):
         return self.A.T.dot(self.b)
 
-    def newParams(self, params, **kwargs):
+    @classmethod
+    def reshape_params(cls, params, params_lens):
+        new_params = np.empty(len(params_lens), dtype=object)
```
jmeyers314, line 966: `dtype=object` always feels awkward to me.  Is there a reason params can't just be a list or list of lists?

jmeyers314 (approved):
LGTM.  Looking forward to the composite models!

I left a few questions/suggestions.

piff/gsobject_model.py
rmjarvis: I believe YAML doesn't recognize tuples, so (0.2, 0.4) gets parsed as a string.  But I'll add a real tuple option as well for people writing this directly from python.

piff/star.py
rmjarvis: Yeah, that's probably a better way to do this.  Not sure why I got myself stuck on it being a numpy.array.

PR #157, opened 2023-07-07
Add the first composite type: Sum

rmjarvis:
This is finally the first composite type that I've been working up to.  Sum is probably a little simpler than Convolve, so I wanted to start with that.  But the other two I have in mind are Convolve and Transform, similar to the corresponding GalSIm object types. 

A Sum PSF is a model made from adding together multiple other PSF types.  Each component can have its own model and interpolation.  The parameters are stored in a list, so each component has its own numpy.array of parameters in the star.fit objects, each of which can be whatever length is required.

The initialization defaults to hsm for the first component and zero for the others, but the user can override this as desired.  I suspect for most use cases, the first component will describe most of the PSF patter, and the later components will fill in the bits that weren't well modeled earlier (e.g. the tree-ring residuals we've seen in DES), so I think this will be a reasonable default initialization.  Certainly we don't want multiple components all using the hsm initialization, since the sum would then be N x the PSF, which wouldn't be a good initial guess for the components.  I suspect that getting good initial guesses for various components will be a big part of making composite PSFs work well, so there may be more work needed eventually on this front.

### DIFF
--- a/piff/sumpsf.py
+++ b/piff/sumpsf.py
@@ -1,0 +1,312 @@
+# Copyright (c) 2016 by Mike Jarvis and the other collaborators on GitHub at
+# https://github.com/rmjarvis/Piff  All rights reserved.
+#
+# Piff is free software: Redistribution and use in source and binary forms
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
+"""
+.. module:: psf
+"""
+
+import numpy as np
+import galsim
+
+from .psf import PSF
+from .util import write_kwargs, read_kwargs
+from .star import Star, StarFit
+from .outliers import Outliers
+
+class SumPSF(PSF):
+    """A PSF class that is the Sum of two or more other PSF types.
+
+    A SumPSF is built from an ordered list of other PSFs.
+
+    When fitting the Sum, the default pattern is that all components after the first one
+    are initialized to zero, and the first component is fit as usual, but just using a
+    single iteration of the fit.  Then the residuals of this model are fit using the second
+    component, and so on.  Once all components are fit, outliers may be rejected, and then
+    the process is iterated.
+
+    This pattern can be tweaked somewhat using the initialization options available to
+    PSF models.  If a component should be initialized to something other than a zero model,
+    then one should explicitly set it.
+
+    :param components:  A list of PSF instances defining the components to be summed.
+    :param outliers:    Optionally, an Outliers instance used to remove outliers.
+                        [default: None]
+    :param chisq_thresh: Change in reduced chisq at which iteration will terminate.
+                        [default: 0.1]
+    :param max_iter:    Maximum number of iterations to try. [default: 30]
+    """
+    _type_name = 'Sum'
+
+    def __init__(self, components, outliers=None, chisq_thresh=0.1, max_iter=30):
+        self.components = components
+        self.outliers = outliers
+        self.chisq_thresh = chisq_thresh
+        self.max_iter = max_iter
+        self.kwargs = {
+            # If components is a list, mark the number of components here for I/O purposed.
+            # But if it's just a number, leave it alone.
+            'components': len(components) if isinstance(components, list) else components,
+            'outliers': 0,
+            'chisq_thresh': self.chisq_thresh,
+            'max_iter': self.max_iter,
+        }
+        self.chisq = 0.
+        self.last_delta_chisq = 0.
+        self.dof = 0
+        self.nremoved = 0
+        self.niter = 0
+        self.set_num(None)
+
+    def set_num(self, num):
+        """If there are multiple components involved in the fit, set the number to use
+        for this model.
+        """
+        if isinstance(self.components, list):
+            # Then building components has been completed.
+
+            # This array keeps track of which num to use for each component.
+            self._nums = np.empty(len(self.components), dtype=int)
+            self._num_components = 0  # It might not be len(self.components) if some of them are
+                                      # in turn composite types.  Figure it out below.
+
+            k1 = 0 if num is None else num
+            for k, comp in enumerate(self.components):
+                self._nums[k] = k1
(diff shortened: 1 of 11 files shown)


piff/sumpsf.py
```diff
@@ -0,0 +1,326 @@
...
+                        new_image -= current_models[kk][i].image
+                modified_star = Star(star.data.withNew(image=new_image), star.fit)
+                modified_stars.append(modified_star)
+
+            # Fit this component
+            new_stars, nremoved1 = comp.single_iteration(modified_stars, logger, convert_func)
+
+            # The fit components in new_stars are what we want, but the data parts are
+            # corrupted by subtracting the other components, which we don't want.
+            # Just copy over the fits.
```
theoschutt, line 197: Something about this wording was very confusing. Possibly-too-wordy suggestion:

> new_stars now has the updated fit components, but the data parts are corrupted by subtracting the other components during fitting. Update the stars by copying the undisturbed data from the original stars and the updated fits from new_stars.

tests/test_sumpsf.py
```diff
@@ -0,0 +1,507 @@
...
+    assert type(psf.components[1].model) is piff.PixelGrid
+    assert type(psf.components[1].interp) is piff.Polynomial
+
+    for i, star in enumerate(psf.stars):
+        print('star',i)
+        target = targets[i]
+        test_star = psf.drawStar(star)
+        # Note: There is no useful test to make of the component parameters.
+        #print('comp0: ',test_star.fit.params[0])
+        #print('comp1: ',test_star.fit.params[1])
```
theoschutt, line 314: What do you mean here?

tests/test_sumpsf.py
```diff
@@ -0,0 +1,507 @@
...
+
+        # Draw should produce something almost identical (modulo bounds).
+        test_image = psf.draw(x=test_star['x'], y=test_star['y'],
+                              stamp_size=config['input']['stamp_size'],
+                              flux=test_star.fit.flux, offset=test_star.fit.center/image.scale)
+        b = test_star.image.bounds & test_image.bounds
+        print('image max diff = ',np.max(np.abs(test_image[b].array - test_star.image[b].array)))
+        np.testing.assert_allclose(test_image[b].array, test_star.image[b].array, atol=1.e-8)
+
+    # Repeat with both components being PixelGrid mdoels.
```
theoschutt, line 274: typo: mdoels -> models

theoschutt (approved):
Some in-line clarification questions, but otherwise looks good to me.

piff/sumpsf.py
```diff
@@ -0,0 +1,326 @@
...
+
+        # Now initialize all the components
+        for k, comp in enumerate(self.components):
+            stars, nremoved1 = comp.initialize_params(stars, logger, default_init=default_init)
+            nremoved += nremoved1
+            # After the first one, set default_init to 'zero'
+            default_init='zero'
+
+        # If any components are degenerate, mark the sum as degenerate.
+        self.degenerate_points = any([comp.degenerate_points for comp in self.components])
```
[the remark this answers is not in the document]
rmjarvis: That's correct.  The composition is of full PSF objects, so they don't interfere with each other.

The only thing this assignment does is to not do outlier rejection on the first iteration if any of the components use basis interpolation.  I'm not sure how important that is, but it seemed sensible.

piff/sumpsf.py
```diff
@@ -0,0 +1,326 @@
...
+        # Fit each component in order
+        for k, comp in enumerate(self.components):
+            logger.info("Starting work on component %d (%s)", k, comp._type_name)
+            # Update stars to subtract current fit from other components.
+            modified_stars = []
+            for i, star in enumerate(stars):
+                new_image = star.image.copy()
+                for kk in range(len(self.components)):
+                    if kk != k:
+                        new_image -= current_models[kk][i].image
```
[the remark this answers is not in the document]
rmjarvis: That's right.  We always subtract off the current best model of all other components, and fit the component being worked on to the residual.
I think this should be pretty stable when applying some tweak to a primary model that is pretty good at describing most of the pattern, but has a residual with a very different interpolation pattern.  E.g. maybe we add on a PixelGrid with much larger extent, but no interpolation, so it just captures the average large-scale residuals around a standard basis-interpolated model over a smaller footprint.  None of my unit tests tried this out, but that's what I mostly have in mind for how the SumPSF would get used in practice.

tests/test_sumpsf.py
rmjarvis: Usually I had been testing the fitted parameters to some known true values.  But for this test, both models are PixelGrid, so the parameters aren't anything I know a priori.  The only thing to test is to use the models to draw images of the total PSF model and see if it matches the data acceptably.

rmjarvis:
@jmeyers314 @aaronroodman 
Would either of you like to take a look at this before merging?  I'll be doing Convolution next if you're rather wait for that as a more realistic composite model.

jmeyers314 (approved):
Hey Mike.  Code all looks good.  I'm wondering if there's room for another practical test though - something like Kolmogorov-to-capture-bulk + PixelGrid-to-mop-up-details, maybe with very different spatial scales?

rmjarvis:
I'm game.  Thoughts on what to use for the truth?  Maybe a GalSim atmospheric screen and then use GP for the interpolation of Kolmogorov for the primary component.  Would that have a usefully interesting residual to mop up?  Or would I need to include some optical component to give it something that wouldn't be well described by Kolmogorov?

jmeyers314:
I'm not sure.  If you keep the simulated exposure time short enough I'm sure an atmsopheric screen true psf would have interesting residuals to mop up.

You might be able to get away with something simpler though, e.g., truth = Moffat with quickly spatially varying beta maybe?
Definitely bonus points though if you get this to work with a GP.

rmjarvis:
Hi Josh, I finally got back to this after 8 months of letting your very good suggestion sit unaddressed.

I added a test using Atmospheric + Optical screens for the PSF.  The model is a Kolmogorov PSF with GP interpolation + a PixelGrid with just a Mean interpolation.

If I don't include the PixelGrid part, here are the images of the residuals (made by the new unit test when running from main):
test_atmpsf_stars2
You can see that there is a roughly consistent pattern to the residuals.  Mostly because the optical screen aberrations were constant over the CCD, so they produces a pretty coherent pattern.

When I include the PixelGrid as a second component in the Sum, here are the residuals:
test_atmpsf_stars

This seems pretty good to me and shows a more realistic use case for how the Sum PSF might be used in practice.

PR #162, opened 2024-04-26
Add ConvolvePSF type

rmjarvis:
This PR adds a new composite PSF type, Convolve.  This is I think the framework into which we can move the work that @aaronroodman et al have done on the OptAtmo type.

Now we can separately specify the two or three things we want to convolve the optical PSF by and the code will separately fit each with their own interpolation scheme, etc.

For now the tests just have a fixed Optical model which is convolved by a Kolmogorov/GP model.  So my next task is to port over the OptAtmo code that fits for the aberrations into something that works in this framework.

### DIFF
--- a/piff/__init__.py
+++ b/piff/__init__.py
@@ -113,6 +113,7 @@ from .optical_model import Optical, optical_templates
 
 # Composite PSFs
 from .sumpsf import SumPSF
+from .convolvepsf import ConvolvePSF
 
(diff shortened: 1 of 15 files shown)


rmjarvis:
Merging this now, since it's been 2 months.  If anyone still wants to review it feel free.  However, I suspect people care more about the applications to Optical + Atmosphere, which are coming next.

PR #165, opened 2024-06-28
Write star.fit.dof to output file

rmjarvis:
@theoschutt noticed that we haven't been writing star.fit.dof to the stars section of the output file.  This was an oversight, which I fix here.


theoschutt (approved):
Looks good to me!

PR #166, opened 2024-06-28
try to switch from numpy/scipy to jax in some part of basis interp

PFLeget:
This Pull Request switch some part of the linear algebra and tensor/vector reshaping from numpy/scipy to JAX in `BasisInterp`. 

While the math are unchanged it accelerate by a factor ~2 the whole computation time while using the current state of the arts in PSF modeling using Piff (`PixelGrid` + `BasisInterp`). Here is one example run at S3DF on one visit of HSC: 

acceleration_jax

And here is a notebook to compare both (classic numpy/scipy vs jax) [on a single ccd from HSC](https://github.com/PFLeget/DM-45055/tree/main).

In order to make pass the test with this branch I had to disable tests in python 3.7 and python 3.8 because I was not able to make JAX work under those version. I found it was ok to disable both as python 3.7 looks to not be supported and python 3.8 will not be supported anymore in October by referring to [this](https://devguide.python.org/versions/).


PFLeget:
I just added a change on the solver. I found it was 2% faster to use Cholesky as a solver. Even if it's not a lot, I am taking it. It does not change the result when I tested it on HSC.

PFLeget:
I got a comment from @jmeyers314 yesterday on performance from JAX on CPU vs number of core used. I did some test and long story short JAX will be faster than numpy/scipy **if the number of core used is greater than one.**  Here are some numbers for reference: 

```
# core = 1
# N_stars = 40
# numpy --> 7.35 sec
# jax --> 9.82 sec

# core = 2
# N_stars = 40
# numpy --> 6.72 sec
# jax --> 4.44 sec

# core = 2
# N_stars = 80
# numpy --> 15.56 sec
# jax --> 9.19 sec

# core = 4 
# N_stars = 40
# numpy --> 6.23 sec
# jax --> 3.04 sec 

# core = 4 
# N_stars = 80
# numpy --> 14.49 sec
# jax --> 6.33 sec
```

piff/basis_interp.py
```diff
@@ -24,6 +24,48 @@
 from .interp import Interp
 from .star import Star
 
+import jax
+# If uncommented, the following line will make the code run in double precision
+# and have identical results as running the original code in double precision.
+# jax.config.update("jax_enable_x64", True)
```
rmjarvis, line 30: I think we always want double precision.  Our matrices are often moderately ill-conditioned, and in my experience, it's rarely worth the headaches of trying to make sure things always converge nicely with single-precision floats.  Even if it works for some set of use cases, one the breaks is usually right around the corner.  So I think we need this line to be uncommented.

piff/basis_interp.py
```diff
@@ -24,6 +24,48 @@
...
+# if jax.config.update("jax_enable_x64", True) it will give the
+# same results as the original code in double precision, but will run
+# slower, but still faster than the numpy/scipy version.
+
+@jit
+def jax_solve(ATA, ATb):
+    # Original code:
+    # dq = scipy.linalg.solve(ATA, ATb, assume_a='pos', check_finite=False)
+    # New code:
+    factor = (jax.scipy.linalg.cholesky(ATA, overwrite_a=True, lower=False), False)
```
rmjarvis, line 46: This is mostly jax's poor API here, but that bare `False` is bad.  It should be named `lower` to be self-documenting.

piff/basis_interp.py
```diff
@@ -24,6 +24,48 @@
...
+    dq = jax.scipy.linalg.cho_solve(factor, ATb, overwrite_b=False)
+    return dq
+
+@jit
+def build_ATA_ATb(alpha, beta, K):
+    ATb = (beta[:, jnp.newaxis] * K).flatten()
+    tmp1 = alpha[:, :, jnp.newaxis] * K
+    tmp2 = K[jnp.newaxis, :, jnp.newaxis, jnp.newaxis] * tmp1[:, jnp.newaxis, :, :]
```
rmjarvis, line 54: Probably should call this `ATA`, not `tmp2`.

piff/basis_interp.py
```diff
@@ -24,6 +24,48 @@
...
+
+@jit
+def vmap_build_ATA_ATb(Ks, alphas, betas):
+    # Use vmap to vectorize build_ATA_ATb across the first dimension of Ks, alphas, and betas
+    vmapped_build_ATA_ATb = vmap(build_ATA_ATb, in_axes=(0, 0, 0))
+    # Get the vectorized results
+    tmp2s, ATbs = vmapped_build_ATA_ATb(alphas, betas, Ks)
+    # Sum the results along the first axis
+    ATb = jnp.sum(ATbs, axis=0)
+    tmp2 = jnp.sum(tmp2s, axis=0)
```
rmjarvis, line 65: ditto

piff/basis_interp.py
```diff
@@ -24,6 +24,48 @@
 from .interp import Interp
 from .star import Star
 
+import jax
```
rmjarvis, line 27: Rather than break py3.7 and 3.8, you should try this and gracefully handle cases where jax is unavailable.
```
try:
    import jax
except ImportError:
    can_use_jax = False
else:
    can_use_jax = True
```
Then maybe check below if the user sets `use_jax=True`, but `can_use_jax is False`, give a warning in the log to tell them it's not supported on their system.

This would also mean that jax would be an optional dependency, not a required one, so users can still run Piff without it.  I'm not sure how heavy a dependency jax is, so it might not be that important, but especially as jax is currently not turned on by default, it doesn't seem like we should required it to be installed.

piff/basis_interp.py
```diff
@@ -249,26 +292,46 @@ def _solve_direct(self, stars, logger):
 
         # Build ATA and ATb by accumulating the chunks for each star as we go.
         nq = np.prod(self.q.shape)
+        logger.info(f'nq = {nq}')
```
rmjarvis, line 295: This is probably a bit baroque for an info.  `logger.debug` would be better.

piff/basis_interp.py
```diff
@@ -354,11 +420,12 @@ class BasisPolynomial(BasisInterp):
                         and is somewhat slower (nearly a factor of 2); however, it is significantly
                         less susceptible to numerical errors from high condition matrices.
                         Therefore, it may be preferred for some use cases. [default: False]
+    :param use_jax:     Use JAX in _solve_direct compared to classic numpy/scipy. [default: False]
```
rmjarvis, line 423:
```suggestion
    :param use_jax:     Use JAX for solving the linear algebra equations rather than numpy/scipy. [default: False]
```

requirements.txt
```diff
@@ -8,3 +8,4 @@ matplotlib>=3.3
 galsim>=2.3
 treegp>=0.6
 threadpoolctl>=3.1
+jax>=0.4
```
rmjarvis, line 11: As mentioned above, I think we can probably remove this from the requirements and make it optional.

rmjarvis (changes requested):
This looks good Pierre-François.  I just have a few comments/suggestions.  Mostly about making the jax stuff optional for now.  But I wouldn't be averse to making JAX required down the line, after we've had a chance to try it out in various contexts to see how the results compare (both on speed and fidelity).

piff/basis_interp.py
PFLeget: Ok sounds good. I will implement it that way.

piff/basis_interp.py
PFLeget: I forgot to remove this line, it was for my personal use when i was debugging. I removed it.

PFLeget:
Ok sounds good. I implemented your comments. Let me know if the change looks good or if it needs some additional changes.

piff/basis_interp.py
```diff
@@ -449,7 +449,6 @@ def __init__(self, order, keys=('u','v'), max_order=None, use_qr=False, use_jax=
         self.use_jax = use_jax
 
         if not CAN_USE_JAX and self.use_jax:
-            logger.warning("JAX is not installed. Switching to numpy/scipy.")
```
rmjarvis: I think this line is worth keeping.  The better solution to the error you were getting is to add
```
logger = galsim.config.LoggerWrapper(logger)
```
before this.

PFLeget:
Just added the warning in the log.

rmjarvis (approved):

rmjarvis:
Thanks PF!  Looks good.  Now we can try it in a few different use cases and get some real work comparisons.  My guess is the speed will depend a fair bit on the kind of hardware being used.

Issue #168, opened 2024-07-11
Serialization for embedding in larger (Rubin) files


TallJimbo:
Rubin's been storing its Piff-generated PSFs via pickle thus far, and we all know that's a terrible idea.  We've had a few vague thoughts about what we might do instead, but some of those are finally firming up and it's past time we did something.

We'd like to be able to store arbitrary Piff PSFs in a backwards-compatible way (future Piff versions can read files written by older versions) inside larger files.  Most of those files will be FITS, but we don't want to assume they all will be.  And while we could have Rubin-side code that introspects Piff objects of the types we actually use and handles the serialization, it be better to do this directly in Piff and more naturally get support for all types of models, interpolants, etc.

Here are some possibilities:

1. One option is Piff's existing FITS serialization, which we probably should have tried to use instead of pickle from the beginning.  In many cases we could give Piff a `fitsio.FITS` object and let it append some additional HDUs.  We'd need to make sure it didn't assume it was writing a full file, but I assume that's either already the case or not hard. And we'd probably need it to report something about what it did write, so that we can have pointers to those HDUs elsewhere in the file.  The big downside of this approach is that it does require FITS, and that means in other contexts (and maybe in some FITS contexts where appending HDUs might be problematic) we'd have to write to FITS and then consider that a binary blob to be stuffed into some other file.

2. If we're going to write a binary blob sometimes (and consider Piff-appended HDUs to be totally opaque in terms of how we document our files for science users in other places), maybe we should just always store Piff PSFs as binary blobs, but make it a Piff method that does that.  The first version could even delegate to the FITS serialization, then run that through gzip or something and prepend a serialization version number for backwards compatibility, and if we wanted to rewrite that without delegating to FITS later it'd be a relatively transparent change (as long as we could still read the FITS version in the future).

[...]

PR #170, opened 2024-07-18
Refactor write/read methods to delegate to new Writer/Reader interface.

TallJimbo:
See #168 for motivation.

This refactors the current FITS read/write code to delegate to new `Reader` and `Writer` classes that have FITS implementations, with a JSON-like alternate implementation envisioned but not implemented.  The FITS implementations should write files that are readable by older versions of Piff (the only change to the outputs is that the Piff version is now written to all of the extension headers, not just some of them), and of course read files written by older versions of Piff.

### DIFF
--- a/piff/psf.py
+++ b/piff/psf.py
@@ -17,12 +17,10 @@
 """
 
 import numpy as np
-import fitsio
 import galsim
-import sys
 
 from .star import Star, StarData
-from .util import write_kwargs, read_kwargs
+from .util import read_kwargs
 
 class PSF(object):
     """The base class for describing a PSF model across a field of view.
@@ -722,34 +720,34 @@ class PSF(object):
         :param file_name:   The name of the file to write to.
         :param logger:      A logger object for logging debug info. [default: None]
         """
+        from .writers import Writer
+
         logger = galsim.config.LoggerWrapper(logger)
         logger.warning("Writing PSF to file %s",file_name)
 
-        with fitsio.FITS(file_name,'rw',clobber=True) as f:
-            self._write(f, 'psf', logger)
+        with Writer.open(file_name) as w:
+            self._write(w, 'psf', logger=logger)
 
-    def _write(self, fits, extname, logger=None):
+    def _write(self, writer, name, logger):
         """This is the function that actually does the work for the write function.
         Composite PSF classes that need to iterate can call this multiple times as needed.
 
-        :param fits:        An open fitsio.FITS object
-        :param extname:     The name of the extension with the psf information.
+        :param writer:      A writer object that encapsulates the serialization format.
+        :param name:        A name to associate with this PSF in the serialized output.
         :param logger:      A logger object for logging debug info.
         """
         from . import __version__ as piff_version
-        if len(fits) == 1:
-            header = {'piff_version': piff_version}
-            fits.write(data=None, header=header)
         psf_type = self._type_name
-        write_kwargs(fits, extname, dict(self.kwargs, type=psf_type, piff_version=piff_version))
-        logger.info("Wrote the basic PSF information to extname %s", extname)
-        if hasattr(self, 'stars'):
-            Star.write(self.stars, fits, extname=extname + '_stars')
-            logger.info("Wrote the PSF stars to extname %s", extname + '_stars')
-        if hasattr(self, 'wcs'):
-            self.writeWCS(fits, extname=extname + '_wcs', logger=logger)
-            logger.info("Wrote the PSF WCS to extname %s", extname + '_wcs')
-        self._finish_write(fits, extname=extname, logger=logger)
+        writer.write_struct(name, dict(self.kwargs, type=psf_type, piff_version=piff_version))
+        logger.info("Wrote the basic PSF information to name %s", writer.get_full_name(name))
+        with writer.nested(name) as w:
+            if hasattr(self, 'stars'):
+                Star._write(self.stars, w, 'stars')
+                logger.info("Wrote the PSF stars to name %s", w.get_full_name('stars'))
+            if hasattr(self, 'wcs'):
+                w.write_wcs_map('wcs', self.wcs, self.pointing)
+                logger.info("Wrote the PSF WCS to name %s", w.get_full_name('wcs'))
+            self._finish_write(w, logger=logger)
 
     @classmethod
     def read(cls, file_name, logger=None):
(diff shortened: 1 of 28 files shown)


piff/psf.py
```diff
@@ -722,34 +719,34 @@ def write(self, file_name, logger=None):
...
         logger = galsim.config.LoggerWrapper(logger)
         logger.warning("Writing PSF to file %s",file_name)
 
-        with fitsio.FITS(file_name,'rw',clobber=True) as f:
-            self._write(f, 'psf', logger)
+        with Writer.open(file_name) as w:
+            self._write(w, 'psf', logger=logger)
 
-    def _write(self, fits, extname, logger=None):
+    def _write(self, writer, name, logger):
```
TallJimbo, line 730: Throughout this ticket I've taken the liberty of just changing the signatures of underscore-prefixed methods (`_write`, `_finish_write`, `_read`, `_finish_read`) but I've kept `read` and `write` the same (while having them delegate to the new methods.  If any of those underscore-prefixed methods shouldn't change because they might be in use outside Piff, I can go back and add new methods instead.

piff/writers.py
```diff
@@ -0,0 +1,257 @@
...
+        raise NotImplementedError()
+
+    def write_wcs_map(self, name, wcs_map, pointing):
+        """Write a regular a map of WCS objects and an optoinal pointing coord.
+
+        :param name:      Name used to save this struct in `Writer.write_array`.
+        :param wcs_map:   A `dict` mapping `int` chipnum to `galsim.BaseWCS`.
+        :param pointing:  A `galsim.CelestialCoord`, or `None`.
+        :param logger:    A logger object for logging debug info.
+        """
```
TallJimbo, line 102: I've moved the WCS reading and writing in the `PSF` class into the implementation of this method and its `Reader` counterpart, instead of leaving them in `PSF` and having them call `write_table` and `read_table`.  That's because I envision reimplementing this specifically for Rubin since we'll be saving the WCSs ourselves elsewhere in the same file as the PSF most of the time, and we might want to just have a string/int pointer to that written by the writer and dereferenced by the reader.

TallJimbo (commented):
One open question is that I'm not sure if I should do something about the `Output` class (I have not yet), which looks like it was intended to do some of the same file-type dispatch as the new `Writer.open` method, but going by code comments never really grew into that role.

TallJimbo:
This is now ready for at least preliminary review, @rmjarvis.  Apparent test failure is actually just hitting the codecov rate limit (sorry!), and it should resolve itself if kicked in a few hours.  Hopefully nothing unexpected turns up in the rest of the build matrix.

piff/readers.py
```diff
@@ -0,0 +1,273 @@
...
+            # numpy scalars don't always satisfy the expected isinstance
+            # checks.
+            if isinstance(v, bytes):
+                struct[k] = v.decode()
+            elif isinstance(v, np.bool_):
+                struct[k] = bool(v)
+            elif isinstance(v, np.integer):
+                struct[k] = int(v)
+            elif isinstance(v, np.floating):
+                struct[k] = float(v)
```
rmjarvis, line 184: This whole bit wasn't in the original, and maybe isn't necessary?  We just returned struct in `read_kwargs`.  So, (a) can you explain a little more why you think this loop is useful? And (b) if it is important to keep, add unit tests for the bytes and bool branches that are currently uncovered.  (We apparently don't have cases of those right now.)

piff/readers.py
```diff
@@ -0,0 +1,273 @@
...
+            metadata.update(self._fits[extname].read_header())
+        return self._fits[extname].read()
+
+    def read_wcs_map(self, name, logger):
+        # Docstring inherited.
+        import base64
+        try:
+            import cPickle as pickle
+        except ImportError:
+            import pickle
```
rmjarvis, line 206: This isn't necessary in py3.

piff/writers.py
```diff
@@ -0,0 +1,257 @@
...
+        the `nested` method.
+        """
+        raise NotImplementedError()
+
+
+class FitsWriter(Writer):
+    """A `Writer` implementation that writes to multiple FITS HDUs.
+
+    This reader is intended to writes files that would be readable Piff before
+    the `Reader` and `Writer` abstractions were added.
```
rmjarvis, line 131: This sentence has multiple errors.  But the following would suffice I think.
"This writer is used to write Piff data to a FITS file."

piff/writers.py
```diff
@@ -0,0 +1,257 @@
...
+    methods of `Reader` each directly correspond to a method of `Writer`.
+
+    All `Reader` and `Writer` methods take a ``name`` argument that is
+    associated with the low-level data structured being saved.  When writing,
+    an object can chose not to write a subobject at all with a given name, and
+    then check to see if the corresponding `Reader` method returns `None`, but
+    a `Reader` cannot be used to see what type of data structure was saved with
+    a given name; if this can change, the write implementation must store this
+    explicitly and use it when invoking the `Reader`.
+    """
```
rmjarvis, line 46: Again, I don't think this base class is necessary or useful.

piff/writers.py
```diff
@@ -0,0 +1,257 @@
...
+    def write_array(self, name, array, metadata=None):
+        # Docstring inherited.
+        if metadata:
+            header = self._header.copy()
+            header.update(metadata)
+        else:
+            header = self._header
+        self._fits.write_image(
+            array, extname=self.get_full_name(name), header=self._header
+        )
```
rmjarvis, line 201: This method seems not to be used?  We don't have any PSF types that write images, so I think for now write_table is all we need.  Although I guess it's not crazy to imagine we might want the option to write images someday.  But it will need some kind of unit test until then, since right now this function doesn't seem to be called by anything.

piff/writers.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def write_wcs_map(self, name, wcs_map, pointing):
+        # Docstring inherited.
+        import base64
+
+        try:
+            import cPickle as pickle
+        except ImportError:
+            import pickle
```
rmjarvis, line 210: ditto, not necessary.

rmjarvis (changes requested):
Thanks Jim.  I have a few comments, but this is broadly what I had in mind.

piff/readers.py
```diff
@@ -0,0 +1,273 @@
...
+
+        :returns:  A context manager that yields a `Reader`.
+        """
+        _, ext = os.path.splitext(file_name)
+        if ext == ".fits" or ext == ".piff":
+            with FitsReader._open(file_name) as reader:
+                yield reader
+            return
+        else:
+            raise NotImplementedError(f"No reader for extension {ext!r}.")
```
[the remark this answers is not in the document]
TallJimbo: I'm happy to drop the `open` method.

I like ABCs as a place to hang docstrings (most of the time derived-class methods can just inherit them) and more generally document the interface that users and implementers are expected to satisfy.  But I'm happy to remove it if it's not the pattern you prefer, and just move the docstrings over to the FITS implementations.

piff/readers.py
TallJimbo: I've been burned in the past by numpy scalar types not being exactly substitutable for their built-in counterparts, and have gotten in the habit of coding defensively around them.  IIRC that's the only rationale for all but the `bytes`, case, which was necessary to have `str` round-trip through the FITS binary table.  But let me see what happens if I drop some or all of these and get back to you with a more precise answer.

piff/writers.py
TallJimbo: Huh, I hadn't realized that.  I think it'd probably be best to just remove it, then, and if we need it someday that will clarify what it should actually look like.

TallJimbo (commented):
I've added some replies and will go do a round of edits.

piff/readers.py
TallJimbo: Ok, I've confirmed that you're right: none of this is necessary, not even the `bytes` -> `str` coercion, and removed it all.  No idea what got me confused on that point.

TallJimbo:
I believe I've addressed all review comments, so this is ready for another pass.

piff/star.py
```diff
@@ -432,19 +432,21 @@ def read_coords_params(cls, fits, extname):
         return coords, params
 
     @classmethod
-    def read(cls, fits, extname):
+    def read(cls, reader, name):
         """Read stars from a FITS file.
```
rmjarvis, line 436:
```suggestion
        """Read a list of stars via an open reader object.
```

rmjarvis (approved):

PR #173, opened 2024-09-11
C++/Eigen implementation of Solve direct for Basis interp

PFLeget:
I pushed a C++ code developed by @natelust  for solving the interpolation in Basis Interp. It goes faster than numpy/scipy/Jax even on a single core. More stars are used for the training and faster it will be compared to numpy/scipy/Jax even on a single core.


src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    // Allocate an array for b, similar to ATA
+    auto ATb = py::array_t<T>(A_shape[1]*N);
+    py::buffer_info ATb_buffer = ATb.request();
+    T __restrict * ATb_data = static_cast<T *>(ATb_buffer.ptr);
+    memset(ATb_data, 0, A_shape[1]*N*sizeof(T));
+
+    // Allocate intermediate arrays that will be reused in the following loops
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> ATA_small(A_shape[1],A_shape[1]);
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> K_cross(N, N);
```
rmjarvis, line 45: If you do end up keeping N as a template, then you're missing an opportunity here.  Use `<T, N, N>` rather than Dynamic.

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+        py::buffer_info K_buffer_info = KList[c].request();
+        T __restrict *K_data = static_cast<T *>(K_buffer_info.ptr);
+
+        // set some variables for sizes
+        ssize_t A_shape_one = A_buffer_info.shape[1];
+        ssize_t A_shape_zero = A_buffer_info.shape[0];
+
+        // calculate the outer product of the K parameters so that it can be re-used in the
+        // following loops
+        Eigen::Map<const Eigen::Vector<T, Eigen::Dynamic>> K_vec(K_data, N);
```
rmjarvis, line 68: Likewise here.  `<T, N>`.

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+        ATA_small.setZero();
+        // since the array will be self-adjoint, use Eigen such that only half the number
+        // of comutations need done.
+        ATA_small.template selfadjointView<Eigen::Upper>().rankUpdate(A_map.transpose());
+
+        Eigen::Map<const Eigen::Matrix<T, Eigen::Dynamic, 1>> b_map(b_data, b_buffer_info.shape[0]);
+        ATb_small = A_map.transpose() * b_map;
+    
+        // Calculate ATb contribution for this list entry
+        for (ssize_t o = 0; o < A_shape_one; o ++) {
```
rmjarvis, line 82: Please don't use o as a variable.  k would be natural here.

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+    }
+
+    // solve Ax=b. If the matrix can be decomposed using llt do that (faster) if it cant be
+    // then use the slower ldlt method
+    Eigen::Map<const Eigen::Vector<T, Eigen::Dynamic>> ATb_map(ATb_data, A_shape[1]*N);
+    auto view = ATA_eig. template selfadjointView<Eigen::Upper>();
+    auto lltDecomp = view.llt();
+    Eigen::Matrix<T, Eigen::Dynamic, 1> result;
```
rmjarvis, line 104: Isn't this just a Vector?

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (lltDecomp.info() == Eigen::ComputationInfo::Success) {
+        result = lltDecomp.solve(ATb_map);
+    } else {
+        auto decomp = view.ldlt();
+        result = decomp.solve(ATb_map);
```
rmjarvis, line 109: If the matrix is singular, this will also fail.  In the python code, we try to do something graceful in that case, switching to SVD.  I suspect we want something similar here.  Or at least a try/catch somewhere to inform the user what happened rather than let through a probably baroque Eigen error message.

rmjarvis (changes requested):
Thanks Nate and Pierre-François.  This is a great addition.  I have some code comments.

But also, could you report some timing differences for some representative use case?  Ideally comparing all 4 solve methods, but at least showing how cpp compares to the default method.

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+        case 6:
+            return _solve_direct_cpp_impl<T, 6>(bList, AList, KList);
+        case 7:
+            return _solve_direct_cpp_impl<T, 7>(bList, AList, KList);
+        case 8:
+            return _solve_direct_cpp_impl<T, 8>(bList, AList, KList);
+        case 9:
+            return _solve_direct_cpp_impl<T, 9>(bList, AList, KList);
+        case 10:
+            return _solve_direct_cpp_impl<T, 10>(bList, AList, KList);
```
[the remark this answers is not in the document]
PFLeget: Ok, so I did a parallel branch of this branch where there is an additional cpp solver without template (`add_cpp_in_BasisInterp_check_template`)

I did some performance test on a single CCD from HSC with ~100 training PSF stars. The default Piff config: 


```
{'type': 'Simple',
 'model': {'type': 'PixelGrid',
  'scale': 0.16839226517851116,
  'size': 25,
  'interp': 'Lanczos(11)'},
 'interp': {'type': 'BasisPolynomial', 'order': 2, 'solver': 'scipy'},
 'outliers': {'type': 'Chisq', 'nsigma': 4.0, 'max_remove': 0.05}}
```

So here this is `case 6:`  and here are the difference in time to compute the PSF solutions for the different solver:

```
Solver = Scipy —> 20.770 seconds 
Solver = C++ (with template, case:6) —> 14.162 seconds
Solver = C++ (no template) —> 16.942 seconds 
```

I switched the order of the interpolation to 3 (`case: 10`, then) and here are the difference in time to compute the PSF solutions. 

```
Solver = Scipy —> 46.795 seconds
Solver = C++ (with template, case:10) —> 29.346 seconds
Solver = C++ (no template) —> 35.622 seconds 
```

So in both case, C++ is faster but this is even faster by using template. My understanding of why this is faster from discussion with Nate, this is coming from this block of lines:  

```
        // Calculate ATA contribution for this list entry
        for (ssize_t j = 0; j < A_shape_one; j++) {
            ssize_t base_pos_j = j*N;
            for (ssize_t i = 0; i < j+1; i++) {
                ssize_t base_pos = i*N;
                ATA_eig.template block<N, N>(base_pos, base_pos_j) += ATA_small(i,j)*K_cross;
                // this is where it was supposed to make the difference because of block assignement.
```

It allows a better implementation in each case at compilation time if I understood in each different case. 

So for the moment I will add case up to 30. Do you think I should add a version without template to be sure there is no error coming from the C++ code in case above 30?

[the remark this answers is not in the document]
rmjarvis: I suspect there are rapidly diminishing returns above N=10.  So I suggest having template cases up that that, and then a non-template for anything larger.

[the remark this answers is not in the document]
rmjarvis: Also, to make your life easier to have both kinds, you can use `Eigen::Dynamic` for when N>10, and also pass `n` as a variable that you use for the loops.  Then `N` can be only used for the Eigen template calls.

piff/basis_interp.py
```diff
@@ -414,22 +439,41 @@ class BasisPolynomial(BasisInterp):
...
+                            and is somewhat slower (nearly a factor of 2); however, it is significantly
+                            less susceptible to numerical errors from high condition matrices.
+                            "jax": Use JAX for solving the linear algebra equations rather than numpy/scipy.
+                            Equivalent to "des" solver therefore, it may be preferred for some use cases. It will
+                            be faster than "des" solver on multi-core cpu / gpu are available.
+                            "cpp": Use C++/eigen for solving the linear algebra equations rather than
+                            numpy/scipy Equivalent to "des" solver, therefore, it may be preferred for some
+                            use cases. On a single core cpu (and more), it will be faster than "des" solver
+                            if the number of training stars is more than ~30.
+    :param cpp_use_float32: When solving with the cpp solver, use float32. [default: False]
```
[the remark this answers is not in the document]
PFLeget: I added some test for the float32 and I added it as an explicit solver. I guess the idea behind it is to use this as a quick (not so accurate) version of the PSF. But might need to be more tested to see how accurate and reliable it is. If you think it should not be here, I can remove it.

src/basic_solver.cpp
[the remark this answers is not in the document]
PFLeget: Ok, so I increased the template to handle up to case 30 and I followed your suggestions when the case is above 30.

piff/basis_interp.py
```diff
@@ -414,22 +438,44 @@ class BasisPolynomial(BasisInterp):
...
+                            be faster than "scipy" solver on multi-core cpu / gpu are available.
+                            "cpp": Use C++/eigen for solving the linear algebra equations rather than
+                            numpy/scipy Equivalent to "scipy" solver, therefore, it may be preferred for some
+                            use cases. On a single core cpu (and more), it will be faster than "scipy" solver
+                            if the number of training stars is more than ~30. With a Piff config using `PixelGrid`
+                            with `size=25` and `interp="Lanczos(11)"` and a second order polynomial for interpolation,
+                            and running on ~O(100) PSFs reserved stars on a 4GB single core CPU, "cpp" solver is 60%
+                            faster than "scipy" solver.
+                            "cpp32": Same as cpp solver but use float32, faster than cpp with a trade for accuracy.
+                            Great power imply great responsability, to use carefully.
```
rmjarvis, line 465: Most of this discussion should be moved up to the text section above the parameters list.  So this can be a short doc that ends at "Solvers available are "scipy", "qr", "jax", "cpp", "cpp32".  See above for details."

piff/basis_interp.py
```diff
@@ -446,13 +492,24 @@ def __init__(self, order, keys=('u','v'), max_order=None, use_qr=False, use_jax=
...
+
+        if solver not in valid_solver:
+            raise ValueError(f"{solver} is not a valid solver. Valid solver are {valid_solver}")
+
+        # To match old API when jax and cpp were not part of solving
+        # basis interp.
+        if use_qr and solver not in ["scipy", "qr"]:
+            raise NotImplementedError(f"use_qr and {solver} are not compatible")
+        if use_qr:
+            self.solver = "qr"
```
rmjarvis, line 508: We should probably deprecate the `use_qr` option.  (I don't think anyone uses it anyway, but better to maintain backwards compatibility.)  So add a warning that `use_qr=True` should now be written `solver='qr'`.

src/basic_solver.cpp
```diff
@@ -0,0 +1,229 @@
...
+        case 27:
+            return _solve_direct_cpp_impl<T, 27>(bList, AList, KList, K_buffer_info.shape[0]);
+        case 28:
+            return _solve_direct_cpp_impl<T, 28>(bList, AList, KList, K_buffer_info.shape[0]);
+        case 29:
+            return _solve_direct_cpp_impl<T, 29>(bList, AList, KList, K_buffer_info.shape[0]);
+        case 30:
+            return _solve_direct_cpp_impl<T, 30>(bList, AList, KList, K_buffer_info.shape[0]);
+        default:
+            return _solve_direct_cpp_impl<T, 31>(bList, AList, KList, K_buffer_info.shape[0]);
```
rmjarvis, line 218: Don't use 31.  Use Eigen::Dynamic.

tests/test_pixel.py
```diff
@@ -1037,28 +1046,26 @@ def test_single_image():
...
-        print("JAX not installed.")
+    configSolver = copy.deepcopy(config)
+    
+    test_star_solver = []
 
-    configJax = copy.deepcopy(config)
-    configJax['psf']['interp']['use_jax'] = True
-    test_star_jax = []
+    solvers = ["scipy", "qr", "jax", "cpp", "cpp32"]
+    rtols = [1e-7, 1e-7, 1e-7, 2e-1]
```
rmjarvis, line 1054: Um.  2e-1 is very high tolerance.  Is the cpp solver really that far wrong?  If so, that's a problem I think.  And should include cpp32 to at least investigate how wrong that is as well.

tests/test_pixel.py
```diff
@@ -1477,8 +1487,6 @@ def test_des2():
     # Basically saying the solution doesn't go completely haywire.
     np.testing.assert_allclose(test_stamp.array, check_stamp.array, rtol=tol_qrp, atol=tol_qrp)
 
-    # Also exercise the SVD fallback to the non-QR method.
-    config['psf']['interp']['use_qr'] = False
```
rmjarvis, line 1481: Why did you remove this?

src/basic_solver.cpp
PFLeget: The problem is that by doing this I got an error at compile time: 

```
/sdf/group/rubin/sw/conda/envs/lsst-scipipe-9.0.0/include/eigen3/Eigen/src/Core/Block.h:131:44: error: static assertion failed: THIS_METHOD_IS_ONLY_FOR_FIXED_SIZE
  131 |       EIGEN_STATIC_ASSERT(RowsAtCompileTime!=Dynamic && ColsAtCompileTime!=Dynamic,THIS_METHOD_IS_ONLY_FOR_FIXED_SIZE)
```

tests/test_pixel.py
PFLeget: Here I compared everything to `"scipy"` solver, so the `rtol=1e-7` is for `"qr"`, `"jax"`, and `"cpp"` compared to `"scipy"` solution. The `rtol=2e-1` is for `"cpp32"` solver that as expected is giving worst results. 

In addition of those test, I had a notebook on the side to make sure that `"scipy" == "jax" == "cpp"`.  `"cpp32"` is not

tests/test_pixel.py
PFLeget: I did not see it was switch to true earlier in the code... I added it back.

tests/test_pixel.py
rmjarvis: I see.  I missed how the +1 was applied.  If cpp32 is only correct at 20% tolerance, I don't think it's worth keeping.  That's really quite bad.

PFLeget: Sounds good to me, I removed it.

src/basic_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::vector<ssize_t> b_shape = b_buffer_info.shape;
+    std::vector<ssize_t> A_shape = A_buffer_info.shape;
+
+    // Allocate and initialize container array for the A transpose A matrix.
+    // For some reason it is faster to allocate a numpy array and then map
+    // it to Eigen than allocating an Eigen array directly
+    // TO DO PF: make sure that A_shape[1] exist. A_shape.at(1)
+    py::array_t<T, py::array::f_style> ATA(std::vector<ssize_t>({A_shape[1]*N, A_shape[1]*N}));
+    py::buffer_info ATA_buffer = ATA.request();
+    T __restrict * ATA_data = static_cast<T *>(ATA_buffer.ptr);
```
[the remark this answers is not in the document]
PFLeget: I was able to synched with Nate and I tested what you suggested in this Branch: `add_cpp_in_BasisInterp_mem_allocation_test`. 

I tried the memory allocation like in GalSim and it does not looks to make a difference in time to compute the PSF solution. Do you want me to keep this like that or to cherry-pick the commit from `add_cpp_in_BasisInterp_mem_allocation_test` and have a more GalSim like memory allocation?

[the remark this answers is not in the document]
rmjarvis: Well, I definitely don't like this version.  If you have something less convoluted that doesn't sacrifice speed, please change to that.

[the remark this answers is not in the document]
rmjarvis: E.g. if a straightforward new works, that is probably fine.  You might not need any of the more complicated versions.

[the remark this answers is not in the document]
PFLeget: I just added it. Does that looks good?

PFLeget:
It looks like I have again same error unrelated to the change I pushed. Locally when I am trying to reproduce it in docker,  all unit test are passing....

rmjarvis:
cf. https://github.com/astropy/astropy/issues/17416

rmjarvis:
I added some more commits.  Mostly style and naming changes.  The substantive changes were:
* Added a workaround for astropy 6.1.5+
* Changed the backup method to FullPivHouseholderQr rather than completeOrthogonalDecomposition.  The latter is overkill and never necessary.
* Use the new aligned memory allocation for ATb as well as ATA
* Use Dynamic for N rather than 31 when the size is not known at compile time.

I'm happy with this now, but I'll leave it for a day in case you want to look over my edits.

rmjarvis (approved):

PFLeget:
Oh thank you so much! Let me run it as it is now and I'll merge it if I don't see any change in run time.

PFLeget:
Ok it looks exactly the same in computation time! So I'll merge it in `main`

PR #174, opened 2024-09-17
Get weight from sky image

rmjarvis:
@esheldon had a use case where there was a sky image, but no weight map.  So this PR adds an option to compute the weight map from the sky image.

There were also a couple places where stats weren't respecting the new flagging interface, so this fixes those.

### DIFF
--- a/piff/star_stats.py
+++ b/piff/star_stats.py
@@ -115,7 +115,8 @@ class StarStats(Stats):
                                      axis=0)
             ave_model_image = np.mean([s.image.array/np.max(s.image.array)
                                         for s in calculated_models
-                                        if s is not None and not s.is_flagged],
+                                        if s is not None and not s.is_flagged
+                                            and np.max(s.image.array) != 0.],
                                       axis=0)
             ave_star_image = galsim.Image(ave_star_image)
             ave_model_image = galsim.Image(ave_model_image)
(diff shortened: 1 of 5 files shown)


piff/star_stats.py
```diff
@@ -115,7 +115,8 @@ def compute(self, psf, stars, logger=None):
                                      axis=0)
             ave_model_image = np.mean([s.image.array/np.max(s.image.array)
                                         for s in calculated_models
-                                        if s is not None and not s.is_flagged],
+                                        if s is not None and not s.is_flagged
+                                            and np.max(s.image.array) != 0.],
```
jmeyers314, line 119: Is it a problem that max(image) == 0 models have survived this far?  Stats feels kinda late in the game to find this.

jmeyers314 (approved):
LGTM.  One question.

piff/star_stats.py
rmjarvis: I think it's super rare.  But there was one case where a model centroid moved by more than the size of the stamp, so the whole model was outside the image.  I don't think we currently have any limit of the amount of centroid shift that is allowed, but it probably makes sense to add something.  This only happened when the model was not really working other reasons anyway, but I didn't want the division by zero in the average.

PR #178, opened 2025-02-25
Only count flags in reflux_star if star is not already flagged

rmjarvis:
@PFLeget noticed a bug in reflux_stars that can cause an (almost) infinite loop.  The problem is that any exception encountered while refluxing gets flagged.  But this happens even if the star is already flagged.  Then in the main fitting routine, it sees what it thinks is a new flagged object and iterates, even if it is the same object failing every time.

The fix is to only count the failure in nremoved if the object is not already flagged.


PFLeget:
I just fixed the problem in treegp for python 3.13 and it is working now. You need to use version 1.3.0 now (I just pushed it).

rmjarvis:
Thanks.  That seems to have worked.

PFLeget:
Do you want me to add a unit test for this?

rmjarvis:
I did add one using mock to force the failure.  But if you have something more realistic, that would be great.

PFLeget:
I asked, and I am allowed to put a sample of the data that made the original problem. I have an example with ~30 stamp of real stars, I would be happy to add it for this, but it will allow me to add other unit test latter on for the cpp solver.

rmjarvis:
Perfect.  Yeah, that would be great.

PFLeget:
I added the test that looks to work on all versions of Python except 3.8. Is it ok to not support python 3.8?

 It looks like python 3.8 is not maintained anymore (https://devguide.python.org/versions/) and when I am trying to have the last version of numpy (numpy>=2.0), it does not support python 3.8.

requirements.txt
```diff
@@ -1,11 +1,11 @@
-numpy>=1.17
+numpy>=2.0
```
rmjarvis, line 1: Don't do this.  This is why py3.8 doesn't work.  We also don't want to require this (at least not yet), since lots of packages don't work yet with it.

requirements.txt
```diff
@@ -1,11 +1,11 @@
...
 scipy>=1.2
 pyyaml>=5.1
 pybind11>=2.2
 lsstdesc.coord>=1.2
-treecorr>=4.3.1
+treecorr>=5.0
```
rmjarvis, line 6: I doubt this is required either.

tests/test_pixel.py
```diff
@@ -1790,6 +1790,50 @@ def test_convert_func():
...
+        },
+        'outliers': {
+            'type': 'Chisq',
+            'nsigma': 4.0,
+            'max_remove': 0.05
+        },
+        'max_iter': 30
+    }
+
+    dic = pickle.load(open('input/testStars.pkl', 'rb'))
```
rmjarvis, line 1824: Probably best to not use pickle.  This can break randomly.  Use the piff.write functionality to write this to a fits file or something.  That should be more stable.  (And not require numpy version to match -- I'm guessing this is why you thought you needed numpy 2.x.)

tests/test_pixel.py
```diff
@@ -1790,6 +1790,50 @@ def test_convert_func():
...
+    wcs = {0: galsim.PixelScale(0.20038369063248057)}
+    pointing = None
+
+    piffResult = piff.PSF.process(piffConfig)
+    piffResult.fit(stars, wcs, pointing)
+
+    if not piffResult.niter == 6:
+        raise ValueError('Maximum number of iterations in this example must be 6.')
+    if not np.shape(piffResult.stars[28].fit.A) == (0, 625):
+        raise ValueError('Centroid failed for this star, expected shape of A is (0,625)')
```
rmjarvis, line 1835: Typically tests use asserts, not raise ValueError.  I.e.
```
assert piffResult.niter == 6
assert piffResult.stars[28].fit.A.shape == (0,625)
```

rmjarvis: Maybe add comments explaining why these are the expected values, which you currently include in your ValueError text.

rmjarvis:
> I see the PR is in direction of v1.5, shall it be towards main or am I missing something?

This is a bugfix.  So we can release it as a bugfix release on the 1.5.x series.  I'll cherry-pick the commits over to main as well ofc.

rmjarvis:
v1.5.1 is released with this bugfix.  It's on pypi so far.  Should be on conda later today.

PR #179, opened 2025-06-24
Make fit_flux=True the default when init=zero

rmjarvis:
This fixes some confusing behavior when using the SumPSF class.  The default initialization for components after the first one is zero, which means they get initialized to have zero flux.  

However, GSObjectModel and PixelGrid both default to using flux=1 all the time.  (All models do really, but these two have the ability to not do so.)  If you don't turn that off, then the second component confusingly keeps setting itself to flux=1, which negates the zero initialization.  This tripped us (@kklaliotis) up today when we were trying to figure out why the SumPSF was behaving so badly on LSSTCam data.

So now, whenever init=zero (either specified explicitly or by default when showing up as a non-first component in a Sum), fit_flux is automatically set to True.  This seems like more intuitive behavior.


PFLeget (approved):
LGTM

PR #180, opened 2025-07-23
Make it possible to exclude cross terms in BasisPolynomial

rmjarvis:
Currently BasisPolynomial has a single number for max_order, which is the same for all cross pairings of two parameters.  So if you want to include uv^2 and u^2v, that's max_order=3.  The problem is when you also have e.g. color or dz - you are stuck with the same max_order for terms like color * v^2.  This is usually not desired.

This PR makes it possible to more carefully specify the max_order you want for different pairs.  max_order is allowed to be a dict whose keys are tuples of parameter names and whose values are the max_order you want for that pairing.

So to have e.g. u,v use 2nd order with cross terms, and also dz use 2nd order with no cross terms with u or v, you would specify: `max_order = { (u,v): 2 }`.  (The missing (u,dz) and (v,dz) pairs are implicitly zero.)

Additionally, we had wanted to try forcing SumPSF to do extra iterations rather than stopping after only 1 or 2.  That wasn't possible.  So I added that parameter to all PSF types, analogous to max_iter.


PFLeget:
I am going to try this branch before approving it, if it sounds ok. I think I can look on DP1 how it impact color correction and on LSSTCam how it change PSF residuals by adding the sensor height on ITL sensor.

PFLeget:
I guess I was referring to something else in DM side. I did not find anything concerning so far.

PFLeget (approved):
LGTM ! Aashay Pai tested during desc hack week It was not changing color fitting to have cross terms. Looks ok to me to be released.

PR #185, opened 2026-03-12
Add badpix_zeros option

rmjarvis:
I have a huge branch called roman that has a bunch of changes I made in order to read/process simulated Roman images made by Chris Hirata.  In order to make the code review feasible, I'm going to parcel out the changes in what I hope are relatively bite-sized pull requests.  This is the first one.

The simulated Roman images that Chris makes don't have an explicit bad-pixel mask.  Rather, they set the image to zero at any pixel that should be considered bad.  This PR adds the option to set badpix_zeros=True to tell Piff to consider any pixel with image value=0 to be a badpixel and thus set the pixel weight to 0.

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -213,6 +213,8 @@ class InputFiles(Input):
                         this gives the name of the file to use. [default: None]
         :badpix_file_name: If the badpix image is in a different file than the main image,
                         this gives the name of the file to use. [default: None]
+        :badpix_zeros:  Whether to treat all zeros in the input image as bad pixels.
+                        [default: False]
 
         :cat_hdu:       The hdu to use in the catalog files. [default: 1]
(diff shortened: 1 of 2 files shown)


aeharbo (approved):
Changes to input.py look good; moving the sky subtraction so that it occurs after the zero-weighting of badpix is the right move

Similarly, the changes to test_input.py seem okay. Usage of the new image_hdu makes sense.

aeharbo (approved):
(Re-doing review, now with the right level of access)
Changes to input.py look good; moving the sky subtraction so that it occurs after the zero-weighting of badpix is the right move

Similarly, the changes to test_input.py seem okay. Usage of the new image_hdu makes sense.

kklaliotis (approved):
this all looks reasonable / simple to me

PR #186, opened 2026-03-12
Enable noise to be specified in FITS header

rmjarvis:
Next Roman processing PR.

The Roman simulated images from Chris don't have weight maps.  Rather the noise level is to be considered a constant, and is given in the FITS header as BKGNDVAR.  The gain is also there as EQVGAIN.

We had the ability to read the gain from the fits header, but not the noise variance.  So this PR adds that capability.

(It also simplifies how the header is passed between functions.  It's now stored as image.header, which is a standard place for it in GalSim, but which we hadn't been using here.)

### DIFF
--- a/piff/input.py
+++ b/piff/input.py
@@ -787,6 +787,12 @@ class InputFiles(Input):
                                "Taking them to be 0.0")
                 weight.array[weight.array < 0] = 0.
         elif noise is not None:
+            try:
+                noise = float(noise)
+            except ValueError:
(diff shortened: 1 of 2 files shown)


aeharbo (approved):
input.py: All type casting, argument, and if statement changes make sense in order to read in noise from the header data

test_input.py: Changes seem good, the assert_almost_equal sets the weight to 1/noise, which is what we want (e.g. 1/ sigma^2)

PR #189, opened 2026-03-12
Enable single image file with multiple hdus

rmjarvis:
Chris packaged the 18 SCAs as a single fits file with multiple hdus.  This obviously makes a lot of sense, but (somewhat surprisingly) it wasn't an input format that Piff could handle yet.  This PR adds that capability.


hirata10 (approved):
The updates make sense, and the test functions work through the appropriate cases.

PR #191, opened 2026-03-12
Enable derived properties, not from input columns

rmjarvis:
The input read can take arbitrary properties from the input catalog and attach them to the stars as "properties".  This is usually sufficient, but it can be useful to compute property values that are not just input columns.  For Roman, this includes computing sca as chipnum+1.  But I think it's broadly useful if you want to do some calculation based on the other input properties read from the catalog and save that as a new property value.  This PR enables all of that.


nihardalal (changes requested):
Some super small changes to submit (see comment on line R1071 and R1080)

nihardalal (approved):

PR #193, opened 2026-03-12
Add min_sep selection option

rmjarvis:
To help catch large centroid drifters before doing the fits, I found it helpful to add a rejection option that rejects objects that are within some distance of some other object in the input catalog.  If the inputs are only stars, no galaxies, this may be of limited usefulness.  But even in that case, it helped reject some stars that would otherwise have eventually been rejected as outliers during the fits.


nihardalal (approved):
Separation seems to be put in correctly with correct and fast implementation. Unit tests are also comprehensive/correct

PR #196, opened 2026-03-25
Add ability to use multiple processes when fitting multiple Roman SCAs

rmjarvis:
This is the next PR for the Roman PSF.  The main feature it adds is to allow multiprocessing during the fits of the 18 SCAs.  

We already had a utility function called run_multi, which we use in a few other places as well.  So most of the changes here are reorganizing the work for each SCA so there is a self-contained job to pass to each worker.

There are a few other minor cleanup items I'm also including:

* When using hsm_size_reject or other things that use hsm, it didn't reject objects for which hsm gave an error.  It should have and now it does.
* In the StarImages output plots, I added the chip number and x,y position to the labels to make it easier to see if there are problems in one or a couple particular SCAs.
* Also for StarImages, if there were fewer stars than requested plots, it didn't work right.  I fixed that.
* I added a warning guard for a newish scipy warning that was being emitted sometimes.
* The number of iterations used wasn't being written to the output file.  This was an oversight, and it's useful to have that information for diagnostics.

### DIFF
--- a/devel/.gitignore
+++ b/devel/.gitignore
@@ -1,0 +1,1 @@
+failed*.fits
--- a/devel/roman/.gitignore
+++ b/devel/roman/.gitignore
@@ -3,3 +3,5 @@ stars_13906_test_v251105.parquet
 Roman_SRR_WFC_Pupil_Mask_Shortwave_2048_reformatted.fits.gz
 *.png
 *.piff
+configs
+output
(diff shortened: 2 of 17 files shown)


aeharbo (approved):
New functions and tests look good. Adding the extra blank lines for clarity on input.py is a good move.

Glad you didn't have to deal with too many re-indexing issues outside of the test functions.

PR #197, opened 2026-04-01
Add two options for how aberrations vary over SCAs

rmjarvis:
The PR adds two features to the RomanOptics PSF class.

1. nominal_interp = "five_points".  The project supplied nominal aberrations on each SCA are supplied at 5 points for each SCA: the four corners plus the center of the SCA.  The default pattern for approximating the PSF in Piff is to just use the 4 corner values and take the PSF at any other position to be a weighted sum of these.  @nihardalal pointed out that we could use all 5 locations where the project gives us information, including the central point instead.  So setting this parameter to five_points rather than bilinear (still the default) does that.

Piff does this a little bit different from what GalSim does when calculating the PSF at an arbitrary location.  GalSim does a bilinear interpolation of the aberrations themselves, and then computes the PSF from that.  Piff computes the 4 or 5 PSFs with the project aberration values and then interpolates the whole PSF.  I don't really have a good feel for which is more accurate.  But the way Piff does it is computationally much more tractable, since constructing the PSF is relatively expensive, so using the interpolated (either bilinear or five_points) PSF is much faster.

In my test exposure, bilinear performed slightly better than five_points, but there wasn't much difference.  We'll see how this goes on the real data eventually.  I think this will at least be something we'll want to be able to test as to which one gives better results.

2. aberration_interp = linear.  In addition to the nominal aberrations, Piff fits for an offset to these.  An "extra_aberrations" array.  The default is to have a single extra_aberrations array for each SCA.  This is aberration_interp="constant", and that is the default.  This PR adds the option to have them be a linear function of x,y across the SCA.  This option gave a worse fit on the test exposure I've been using, so constant is still the default, but we will definitely want to have this option available to us when testing the models on real data.  The issue is that there are 3x as many degrees of freedom with this fit, and the data aren't strongly constraining.  (The prior needs to be quite tight to avoid it going off the rails.)  But with real data, we may be able to have very tight priors on linear patterns based on an ensemble of exposures, in which case this option may become more viable.

### DIFF
--- a/tests/test_roman.py
+++ b/tests/test_roman.py
@@ -58,6 +58,7 @@ def test_roman_optics():
         assert psf.fit_center is False
         assert psf.include_model_centroid is False
         assert psf.model.aberration_interp == 'constant'
+        assert psf.model.nominal_interp == 'bilinear'
         assert psf.interp.per_sca is True
 
         # Global mode uses one aberration vector for full focal plane.
@@ -196,6 +197,10 @@ def test_roman_optics():
         RomanOpticalModel(filter='H158', chromatic=False, max_zernike=6, aberration_interp='bad')
     assert "must be one of" in str(err.value)
 
+    with pytest.raises(ValueError) as err:
+        RomanOpticalModel(filter='H158', chromatic=False, max_zernike=6, nominal_interp='bad')
+    assert "must be one of" in str(err.value)
+
 @timer
 def test_roman_corner_cache():
     """Verify corner-profile caching reuses one 4-corner set for same SCA and params.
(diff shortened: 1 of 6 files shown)


rmjarvis:
By the way, no need to do a careful code review of the two diagnostic scripts in the devel/roman directory.  These are scripts that Codex wrote for evaluating the quality of the model solutions for the piff.yaml runs there.

tests/test_roman.py
```diff
@@ -226,16 +231,90 @@ def test_roman_corner_cache():
...
+        stars5, _ = psf5.initialize_params(stars, logger=logger)
+        psf5.drawStar(stars5[0])
+        params5 = stars5[0].fit.params
+        profiles5 = psf5.model._get_sample_profiles(stars5[0], params5, cache=True)
+        assert len(profiles5) == 5
+        assert len(psf5.model._corner_cache[5][2]) == 5
+
+
+@timer
+def test_five_point_weights():
```
nihardalal: This test looks fine to me!

nihardalal (approved):
Looks good to go. I think you caught all the variable renames, and the unit tests seem pretty good to me as well. As I mentioned in the in-line reviews, we should definitely do more testing of five-points vs bilinear (either on other exposures or real data)

PR #198, opened 2026-04-05
Make the RomanOptics chromatic=True option more effective and more efficient.

rmjarvis:
This PR is focused on the implementation of the chromatic=True option for RomanOptics.  It includes the following pieces:

* Added the bandpass specification in input, rather than as a detail of the RomanOptics PSF.  
* If the bandpass is a galsim.roman version, then RomanOptics class can get the filter name from that.  If not (e.g. if we want to specify a different version that is more accurate perhaps), then filter_name can be specified manually for use in the getPSF call.
* The input SED is immediately multiplied by the bandpass, and that combination is thinned directly, rather than only thinning the SED and still using the unthinned Bandpass, which was still quite slow.
* Since we've already multiplied by the filter throughput, the draw command needs to use a flat version of the bandpass with just the end points, and throughput = 1 in between.  This is called flat_bandpass in the code.
* I added a more aggressive kind of thinning, which specifies a maximum number of samples to use in the SED.  E.g. sed_max_samples=3 keeps only 3 points, so piecewise linear with 2 segments.  I think given how uncertain our SED knowledge will be in practice, this will be acceptable on real data.  Either sed_tol or sed_max_samples or both are available for us to try out on the data.
* Fixed a few bugs that turned up related to using ChromaticObjects in various places.
* Fixed an error in both RomanOpticalModel and OpticalModel.  They should both have _centered=True, not False.  With _centered=False, they don't re-centroid properly with uncertain star positions.  This hadn't turned up, since both of them are usually used in combination with other components (e.g. in a SumPSF), which allows for recentroiding.  It turned up in one of my tests here, but the error is unrelated to the chromatic option.

The config file in devel/roman using chromatic=True is about 5-6 times slower than the achromatic version.  This is with just a linear approximation to the SED (sed_max_samples=2) -- it's almost double that with 3 samples.  This seems reasonable enough given the extra work intrinsic to drawing chromatically, but there might be further optimizations available  to try.  E.g. I haven't tried using photon shooting (for either one), which might be worth trying despite the fact that it results in noisy images (which makes the empirical derivatives tricky, but there might be ways around that).  I also don't have a good feel for how much the different SED approximations matter in practice, since we haven't tested yet with realistic SEDs.

I'll also reiterate that even with chromatic=False, the final PSF model *is* chromatic.  Just the fitting is done achromatically with the PSF being evaluated at just a single wavelength (the bandpass effective wavelength).  I'm open to adjustments to the naming of this parameter, since it's potentially confusing as it currently stands.


piff/galsim_patch.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if bandpass is not None:
+            # Convert the thinned product back into an SED by dividing out the bandpass
+            # wherever the throughput is non-zero.
+            in_band = np.logical_and(newx >= native_blue_limit, newx <= native_red_limit)
+            tp_native = np.zeros_like(newx, dtype=float)
+            bp_wave_native = _bandpass_native_waves(newx[in_band])
+            tp_native[in_band] = bandpass._tp(bp_wave_native) / bandpass.wave_factor
+            nz = tp_native != 0.  # Don't divide by 0.
+            assert np.all(newf[~nz] == 0.)
```
HyeongHan: It's an temporary compatibility patch, but for clarity an error message can be added something like "thinned sed has non-zero values where throughput is zero"

piff/galsim_patch.py
```diff
@@ -0,0 +1,92 @@
...
+            # want to preserve for this observation.
+            tp_native = np.zeros_like(rest_wave_native, dtype=float)
+            bp_wave_native = _bandpass_native_waves(rest_wave_native[in_band])
+            tp_native[in_band] = bandpass._tp(bp_wave_native) / bandpass.wave_factor
+            spec_native *= tp_native
+
+        # Note that this is thinning in native units, not nm and photons/nm.
+        interpolant = (self.interpolant if not isinstance(self._spec, LookupTable)
+                       else self._spec.interpolant)
+        newx, newf = utilities.thin_tabulated_values(
```
HyeongHan: it can be renamed as thinned_lambda thinned_flux but if this module is going to be removed anyway please ignore this comment

piff/psf.py
```diff
@@ -391,8 +408,10 @@ def fit(self, stars, wcs, pointing, logger=None, convert_funcs=None, draw_method
         from .config import LoggerWrapper
         logger = LoggerWrapper(logger)
 
-        self.wcs = wcs
-        self.pointing = pointing
+        if self.wcs is None:
+            logger.error("WARNING: wcs and pointing should now be given in process, not fit.")
```
HyeongHan: If it's warning, shouldn't it be logger.warning?
If it's an error, the message should be ERROR.
self.pointing is None condition also be added?

piff/convolvepsf.py
```diff
@@ -106,6 +106,12 @@ def interp_property_names(self):
             names.update(c.interp_property_names)
         return names
 
+    def set_context(self, wcs, pointing, bandpass):
+        super().set_context(wcs, pointing, bandpass)
+        if isinstance(self.components, list):
```
HyeongHan, line 111: if self.components is not None: can be more straight forward.

piff/galsim_patch.py
rmjarvis: Comments on this file should probably be migrated to https://github.com/GalSim-developers/GalSim/pull/1355.
(I no longer have this patch in the PR.)  But also, I don't think it's possible for this assert to trigger.  That's why it's an assert, not a normal error.

piff/psf.py
rmjarvis: All such messages in Piff are done through the logger mechanism, not python's warnings module.  Given that, logger.error vs logger.warning dictates what verbosity level it shows up.  logger.error's are always written, which I think makes sense for deprecation warnings.  The user should fix it.  Warnings that are merely about something going a little weird in the processing, but might be ok are logger.warning.

HyeongHan (approved):

piff/convolvepsf.py
rmjarvis: None wouldn't work.  It's either a list or an int.  The latter is when it is still in the process of being read in from a file, but _finish_read hasn't been called yet.  The other check for whether the read is complete uses this, so I'm inclined to keep it this way for consistency.

piff/roman/roman_psf.py
```diff
@@ -528,13 +541,13 @@ def getProfile(self, params=None, star=None, cache=True):
         sample_profiles = self._get_sample_profiles(star, params, cache=cache)
         return self._interpolate_samples(star, sample_profiles)
 
-    def _draw_model_image_from_samples(self, star, sample_profiles, convert_func=None):
+    def _draw_model_image_from_samples(self, star, sample_profiles, convert_func):
```
nihardalal, line 544: Why is `convert_func` a mandatory argument now?

piff/roman/roman_psf.py
```diff
@@ -223,6 +219,32 @@ def initialize_iteration(self):
...
+            elif getattr(bandpass, 'name', None) is not None:
+                self.filter_name = bandpass.name
+            else:
+                raise ValueError(
+                    "RomanOptics requires filter_name when bandpass has no name attribute"
+                )
+        self.bandpass = bandpass
+        self.flat_bandpass = make_flat(bandpass)
+
+    def _require_bandpass(self):
```
nihardalal, line 235: Maybe consider renaming this method to `_get_bandpass`? This is just a matter of preference, but require suggests a bool output to me, and we use this more like a get

piff/roman/roman_psf.py
```diff
@@ -653,9 +672,14 @@ def set_num(self, num):
         # Only call set_num if they are actually built.
         if isinstance(self.model, Model):
             self.model.set_num(num)
-        if isinstance(self.interp, Interp):
```
nihardalal, line 656: Isn't the `interp` instance also potentially not real during the read?

piff/input.py
```diff
@@ -1124,9 +1166,106 @@ def _parse_flux_type(flux_type):
...
+        A[wave == sample_wave[-1], -1] = 1.0
+        # Weight by the local wavelength interval so the fit roughly respects the integral
+        # over wavelength rather than just pointwise errors on a nonuniform grid.
+        sqrt_w = np.sqrt(InputFiles._trapezoid_weights(wave))
+        Aw = A * sqrt_w[:, np.newaxis]
+        fw = flux * sqrt_w
+        g = np.linalg.lstsq(Aw, fw, rcond=None)[0]
+
+        # g is now the vector that gives the best least squares approximation f'.
+        return g
```
nihardalal, line 1219: Makes sense to me!

piff/input.py
```diff
@@ -1124,9 +1166,106 @@ def _parse_flux_type(flux_type):
...
+
+            # Use weighted residuals so the chosen samples target the integral-relevant
+            # approximation error, not just the largest pointwise deviation on the candidate grid.
+            resid = np.abs(target - approx) * np.sqrt(cand_weights)
+
+            # Add the item with the highest weighted residual
+            resid[selected] = -1.0  # Don't reselect something we already have.
+            bisect.insort(selected, np.argmax(resid))
+
+        return candidate_wave[selected]
```
nihardalal, line 1262: Also looks good to me!

piff/input.py
```diff
@@ -1166,6 +1305,46 @@ def _read_sed_file(sed_file_name, sed_wave_type, sed_flux_type,
...
+        sed = sed * bandpass
+
+        # This means that the bandpass for drawing needs to be a flat bandpass with the same
+        # limits as the original, which we call flat_bandpass.
+        flat_bandpass = make_flat(bandpass)
+
+        wave = np.array(sed.wave_list, dtype=float)
+        if len(wave) == 0:
+            # If the sed doesn't have a wave_list, we can't thin it.  (This is very rare!)
+            sed = sed.withFlux(1.0, flat_bandpass)
```
nihardalal, line 1321: I think this could potentially be a failure mode that merits some extra documentation? I can imagine us accidentally having SEDs saved in an incorrect format and then this being some weird behavior that shows up (but I might be overthinking)

tests/test_input.py
nihardalal, line 1: Tests look good to me here!

tests/test_roman.py
```diff
@@ -664,6 +712,110 @@ def test_fit_many_nproc():
...
+        psf.fit(stars, logger=logger)
+        assert psf.niter == 1
+        assert psf.nremoved == 0
+        for star in psf.stars:
+            print(star.fit.flux, star.fit.center)
+            assert flux/2 < star.fit.flux < flux*2  # < factor of 2 change from reflux.
+            assert 0 < np.abs(star.fit.center[0]) < 0.05  # centroid shift should be small.
+            assert 0 < np.abs(star.fit.center[1]) < 0.05
+
+
```
nihardalal, line 818: Tests looks good and comprehensive

nihardalal (approved):
Looks good from my end - I think I have a couple questions/comments here and there, feel free to answer as/if necessary

piff/roman/roman_psf.py
rmjarvis: I generally prefer my backend functions (I.e. not ones that users would use) to not use default arguments.  That way when I change anything about them, I immediately catch any uses where I forgot to update to a new signature, rather than silently using a default value that isn't appropriate.  So this should not have used a default =None in the first place.

piff/roman/roman_psf.py
rmjarvis: The only thing this does is validate that the PSF object has a bandpass set (which happens upstream of the Roman-specific processing).  Returning that value is secondary.  Maybe the better change is just to remove the return value.

piff/roman/roman_psf.py
rmjarvis: Yes, but they are real or not real in tandem.  So I simplified this to a single check.  If Model is real, so is interp, and they can both do their set_num now.  Otherwise they are both fake and will be dealt with by the read function.

piff/roman/roman_psf.py
nihardalal: I think it is called in a place where it does actually return the value, so feel free to leave as is.

piff/input.py
rmjarvis: It's pretty rare, but I think it's not something that would merit an error.  It requires both the SED and the bandpass to be lambda expressions, rather than lookup tables read from a file.  This is technically allowed, although it would require a user doing some very specific special case work.  There is a test that covers this branch (https://github.com/rmjarvis/Piff/pull/198/changes#diff-4650d01edc49b02fbf29d980f24442b85e599827c64340d1ec43820d757e511eR2246), but I'm not really worried about this happening by accident in realistic scenarios.

PR #205, opened 2026-08-11
Fix TwoDHist plot and GSObject fitting

rmjarvis:
This PR has two extremely small fixes that don't have anything to do with each other, besides that they were prompted by runs on Roman coadd images by Brenna (@WellsBN).

1. We noticed when looking at the TwoDHist outputs that the T_model panel was actually just a copy of the T panel.  This was a straightforward copy/paste bug in the code, which is fixed here.
2. When fitting the Gaussian profile, she was sometimes getting crashes that seemed to be out-of-memory errors, from extremely large memory allocations.  I'm pretty sure this is because scipy was sometimes trying bad parameter combinations in the non-linear fitting.  This used to work fine, since GalSim would throw an exception when the FFT was too large, and Piff would catch those errors and call the proposed parameter bad.  But now the default behavior is to just warn for that rather than error.  The fix is to just change GalSim's default back to errors, which is a simple global variable edit.

### DIFF
--- a/piff/gsobject_model.py
+++ b/piff/gsobject_model.py
@@ -26,8 +26,8 @@ from .util import estimate_cov_from_jac
 
 # The scipy iterations can get into pathalogical places in parameter space that trigger
 # enormous FFTs. Rather than crash the machine, let GalSim raise an error rather than just
-# a warning for these. The fitting code catches that and calls that parameter try to
-# have a terrible chisq, so it gets rejected by scipy.
+# a warning for these. The fitting code catches that and gives that parameter attempt
+# very large residuals, so it gets rejected by scipy.
 galsim.errors.raise_fft_size_error = True
 
 class GSObjectModel(Model):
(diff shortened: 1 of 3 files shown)


nihardalal (approved):
looks good to me!

piff/gsobject_model.py
```diff
@@ -24,6 +24,11 @@
 from .star import Star
 from .util import estimate_cov_from_jac
 
+# The scipy iterations can get into pathalogical places in parameter space that trigger
+# enormous FFTs. Rather than crash the machine, let GalSim raise an error rather than just
+# a warning for these. The fitting code catches that and calls that parameter try to
+# have a terrible chisq, so it gets rejected by scipy.
+galsim.errors.raise_fft_size_error = True
```
sidneymau, line 31: Is there a better way to identify pathological places of parameter space than just "we don't think large FFTs should be necessary for non-pathological places"?

rmjarvis: The only pathological thing we explicitly check for is |g| >= 1.  Otherwise, we let scipy explore wherever it wants.  And then when it gets somewhere that GalSim says would use too much memory, we rejected it with residuals equal to 1.e100.
cf. https://github.com/rmjarvis/Piff/blob/releases/1.6/piff/gsobject_model.py#L190

